PR #12, opened 2017-03-03
xcodebuild settings

godrei:


xcodebuild/show_build_settings.go
```diff
@@ -0,0 +1,88 @@
...
+	return command.GetCmd()
+}
+
+func parseBuildSettings(out string) (map[string]string, error) {
+	settings := map[string]string{}
+
+	scanner := bufio.NewScanner(strings.NewReader(out))
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		split := strings.Split(line, "=")
```
trapacska: if split := strings.Split(line, "="); len(split) == 2

PR #46, opened 2017-11-15
added further mac compatibility

trapacska:


profileutil/info_model.go
```diff
@@ -112,7 +121,13 @@ func NewProvisioningProfileInfoFromFile(pth string) (ProvisioningProfileInfoMode
 		return ProvisioningProfileInfoModel{}, err
 	}
 	if provisioningProfile != nil {
-		return NewProvisioningProfileInfo(*provisioningProfile)
+		profileType := []ProfileType{}
```
godrei: why is this an array?

PR #54, opened 2017-11-20
Printer

kdobmayer:


certificateutil/info_model.go
```diff
@@ -22,6 +25,23 @@ type CertificateInfoModel struct {
 	certificate x509.Certificate
 }
 
+// String ...
+func (info CertificateInfoModel) String() string {
+	printable := map[string]interface{}{}
+	printable["name"] = info.CommonName
+	printable["serial"] = info.Serial
+	printable["team"] = fmt.Sprintf("%s (%s)", info.TeamName, info.TeamID)
+	printable["expire"] = info.EndDate.String()
```
godrei: add errors filed - the only one error is when the cert is expired

PR #60, opened 2018-04-13
CodeSignGroup interface

BirmacherAkos:


export/codeSignGroup.go
```diff
@@ -0,0 +1,13 @@
+package export
```
godrei: do not use camelCase file name, eighter name it: `codesign_group.go` or `group.go`

export/codeSignGroup.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/bitrise-tools/go-xcode/certificateutil"
+	"github.com/bitrise-tools/go-xcode/profileutil"
+)
+
+// CodeSignGroup ...
+type CodeSignGroup interface {
+	GetCertificate() certificateutil.CertificateInfoModel
```
godrei: no need for Get prefixes

PR #63, opened 2018-07-02
Archive renamed to Build and changed to be more general (archive, build, analyze).

BirmacherAkos:


xcodebuild/build.go
```diff
@@ -30,8 +29,18 @@ xcodebuild -workspace <workspacename> \
...
-type ArchiveCommandModel struct {
+// const ...
+const (
+	Archive BuildCommand = "archive"
+	Build   BuildCommand = "build"
+	Analyze BuildCommand = "analyze"
+)
+
+// BuildCommand can be {archive, build, analyze}
+type BuildCommand string
```
godrei: - do not mention the possible values in the comment (they are listed in the `const` block)
- i would call this type as: `Command / Action` and the values: `ArchiveAction, BuildAction ...`

xcodebuild -h:

`xcodebuild [-project <projectname>] [[-target <targetname>]...|-alltargets] [-configuration <configurationname>] [-arch <architecture>]... [-sdk [<sdkname>|<sdkpath>]] [-showBuildSettings] [<buildsetting>=<value>]... !!![<buildaction>]!!!...`

-> it is called: build action

xcodebuild/build.go
```diff
@@ -30,8 +29,18 @@ xcodebuild -workspace <workspacename> \
...
+
+// BuildCommandModel ...
+type BuildCommandModel struct {
```
godrei: simply Command? it would look nice when it is called: `xcodebuild.Command`

xcodebuild/build.go
```diff
@@ -42,78 +51,96 @@ type ArchiveCommandModel struct {
...
+	buildCommand BuildCommand
 }
 
-// NewArchiveCommand ...
-func NewArchiveCommand(projectPath string, isWorkspace bool) *ArchiveCommandModel {
-	return &ArchiveCommandModel{
-		projectPath: projectPath,
-		isWorkspace: isWorkspace,
+// NewBuildCommand ...
+func NewBuildCommand(projectPath string, buildCommand BuildCommand, isWorkspace bool) *BuildCommandModel {
```
godrei: - NewCommand ? 
- keep the `isWorkspace` close to the `projectPath ` since they are related

xcodebuild/build.go
BirmacherAkos: I agree but it will conflict with the xcodebuild.go/CommandModel interface{}.
I'm thinking about it.

xcodebuild/build.go
```diff
@@ -30,8 +29,18 @@ xcodebuild -workspace <workspacename> \
...
+	ArchiveAction Action = "archiveAction"
+	BuildAction   Action = "buildAction"
+	AnalyzeAction Action = "analyzeAction"
+)
+
+// Action ...
+type Action string
+
+// Command ...
+type Command struct {
```
godrei: CommandBuilder

PR #64, opened 2018-07-04
tvOS profile prefix

BirmacherAkos:


profileutil/info_model_test.go
```diff
@@ -6,13 +6,18 @@ import "github.com/stretchr/testify/require"
 func TestIsXcodeManaged(t *testing.T) {
 	xcodeManagedNames := []string{
```
godrei: pls add test for macos managed profiles as well

PR #66, opened 2018-08-27
disableProvisioningProfileSpecifier and destination added:

BirmacherAkos:
If the disableProvisioningProfileSpecifier is true it will add an empty "PROVISIONING_PROFILE_SPECIFIER=" to the build options.

### DIFF
--- a/xcodebuild/build.go
+++ b/xcodebuild/build.go
@@ -48,12 +48,11 @@ type CommandBuilder struct {
 	destination   string
 
 	// buildsetting
-	forceDevelopmentTeam                string
-	forceProvisioningProfileSpecifier   string
-	forceProvisioningProfile            string
-	forceCodeSignIdentity               string
-	disableCodesign                     bool
-	disableProvisioningProfileSpecifier bool
+	forceDevelopmentTeam              string
+	forceProvisioningProfileSpecifier string
+	forceProvisioningProfile          string
+	forceCodeSignIdentity             string
+	disableCodesign                   bool
 
 	// buildaction
 	customBuildActions []string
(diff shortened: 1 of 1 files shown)


xcodebuild/build.go
```diff
@@ -45,13 +45,15 @@ type CommandBuilder struct {
...
-	forceProvisioningProfileSpecifier string
-	forceProvisioningProfile          string
-	forceCodeSignIdentity             string
-	disableCodesign                   bool
+	forceDevelopmentTeam                string
+	forceProvisioningProfileSpecifier   string
+	forceProvisioningProfile            string
+	forceCodeSignIdentity               string
+	disableCodesign                     bool
+	disableProvisioningProfileSpecifier bool
```
godrei: why just the specifier? why do not you use forceProvisioningProfileSpecifier with empty value?

xcodebuild/build.go
```diff
@@ -154,26 +168,38 @@ func (c *CommandBuilder) cmdSlice() []string {
...
+		slice = append(slice, fmt.Sprintf("PROVISIONING_PROFILE_SPECIFIER="))
 	}
+
 	if c.forceProvisioningProfile != "" {
 		slice = append(slice, fmt.Sprintf("PROVISIONING_PROFILE=%s", c.forceProvisioningProfile))
 	}
+
 	if c.forceCodeSignIdentity != "" {
 		slice = append(slice, fmt.Sprintf("CODE_SIGN_IDENTITY=%s", c.forceCodeSignIdentity))
 	} else if c.disableCodesign {
```
godrei: in case of `disableCodesign ` i think we should also: `PROVISIONING_PROFILE_SPECIFIER="" & PROVISIONING_PROFILE =""`

xcodebuild/build.go
godrei: use forceProvisioningProfileSpecifier, forceProvisioningProfile and forceCodeSignIdentity pointers to distinguish between empty string and not set cases

PR #68, opened 2018-10-05
fix: add the platform switch to the GetExportMethod()

BirmacherAkos:
It will distinguish the iOS and the macOS platform.


profileutil/plist_data.go
```diff
@@ -104,12 +104,19 @@ func (profile PlistData) GetExportMethod(profileType ProfileType) exportoptions.
 
 	entitlements, ok := data.GetMapStringInterface("Entitlements")
 	if ok {
-		if allow, ok := entitlements.GetBool("get-task-allow"); ok && allow {
-			return exportoptions.MethodDevelopment
+		platform, _ := data.GetStringArray("Platform")
```
godrei: either use the function argument `profileType` or determine it based on the profile content.
i vote to determine based on the profile content and remove the `profileType ProfileType` argument.

PR #71, opened 2019-01-07
Adding support for float64 and string map array.

lpusok:


plistutil/plistutil.go
```diff
@@ -182,3 +197,35 @@ func (data PlistData) GetMapStringInterface(forKey string) (PlistData, bool) {
...
 	return nil, false
 }
+
+func castToMapStringInterfaceArray(obj interface{}) ([]PlistData, error) {
+	array, ok := obj.([]interface{})
+	if !ok {
+		return nil, errors.New("failed to cast to []interface{}")
+	}
+
+	casted := []PlistData{}
```
BirmacherAkos: I think you should use the 
`var casted []PlistData` instead

lpusok: fixed

PR #74, opened 2019-02-21
Fixing xcarchive app lookup

lpusok:


xcarchive/ios.go
```diff
@@ -109,7 +110,7 @@ func NewIosWatchApplication(path string) (IosWatchApplication, error) {
 	}
 
 	extensions := []IosExtension{}
-	pattern := filepath.Join(path, "PlugIns/*.appex")
+	pattern := filepath.Join(regexp.QuoteMeta(path), "PlugIns/*.appex")
```
trapacska: applying `regexp.QuoteMeta` on `path` or wiring in a `patternPrefix` var instead of using `regexp.QuoteMeta` everywhere is a workable way?

xcarchive/ios.go
```diff
@@ -210,17 +211,19 @@ func NewIosArchive(path string) (IosArchive, error) {
...
-		}
-
 		appPath := ""
-		if len(pths) > 0 {
-			appPath = pths[0]
+		if appRelativePathToProducts, found := applicationFromPlist(infoPlist); found {
+			appPath = filepath.Join(path, "Products", appRelativePathToProducts)
 		} else {
-			return IosArchive{}, fmt.Errorf("failed to find main app, using pattern: %s", pattern)
+			var err error
```
trapacska: this is only required because you are using scope  bumps, otherwise here you should have an existing instance of the err var. I think it is not a good idea to use scope bump just for readability. These blocks works like a miniature inline functionality, used in cases when you need to use same variable names in the same place.

https://medium.com/golangspec/scopes-in-go-a6042bb4298c

xcarchive/ios.go
```diff
@@ -237,6 +240,26 @@ func NewIosArchive(path string) (IosArchive, error) {
 	}, nil
 }
 
+func applicationFromPlist(InfoPlist plistutil.PlistData) (string, bool) {
+	properties, found := InfoPlist.GetMapStringInterface("ApplicationProperties")
```
trapacska: please move this to the if

PR #75, opened 2019-02-25
Bugfix/xcarchive app lookup

lpusok:
WIll lookup .app in xcarchive based on info.plist contents, if no path is provided will lookup using pattern matching.
On go version 1.8.3 filepath.Glob does not match escaped `.` character (e.g test.xcodeproj) that's why a custom helper escaper function is used instead of regexp.QuotaMeta. Escape character is escaped inside go strings.


xcarchive/glob.go
```diff
@@ -0,0 +1,12 @@
+package xcarchive
+
+func escapeGlobPath(path string) string {
+	var escaped string
+	for _, ch := range path {
+		if ch == '[' || ch == '*' || ch == '?' || ch == '\\' {
```
BirmacherAkos: I think the `]` is missing from the list.

BirmacherAkos:
- What was the issue with the `regexp.QuoteMeta()` solution?
- Why do we need to use double `\\` escape characters?

xcarchive/glob.go
lpusok: if '[' is escaped ']' will not count as control character, based on https://golang.org/src/path/match.go.

PR #77, opened 2019-04-09
Disable index

BirmacherAkos:


xcodebuild/build_test.go
```diff
@@ -0,0 +1,458 @@
...
+		disableIndexWhileBuilding         bool
+		customBuildActions                []string
+		archivePath                       string
+		customOptions                     []string
+		sdk                               string
+		action                            Action
+		want                              []string
+	}{
+		{
+			name:                              "build simulator",
```
godrei: hmm i do not really see a pattern in the tests / or what are we checking with the tests.
i think you should create exclusive tests (test only one thing per test case):
- xcodeproj needs to be added with `-project`, while workspace with `-workspace` flag
- disableCodesign should set 4 build envs
- disableCodesign means any force codesign field will be shadowed
- any of the flag - value options (foe example destination)
- disableIndexWhileBuilding adds the right env
- archive action allows the archive path to be set
- c.customBuildActions and c.customOptions are added to the command.

In each test case just define the property what the test checks (in some cases it make sense to combine cases, like customBuildActions + customOptions)

BirmacherAkos: Fixed in:
https://github.com/bitrise-io/go-xcode/pull/77/commits/7ab6bdff8220583ed855daffeb54b7b4fb889052

And in:
https://github.com/bitrise-io/go-xcode/pull/77/commits/33590e673b6323826dd2a7794c951b861d8113eb

PR #83, opened 2019-07-03
Tool 574 support multiple identities in p12

lpusok:


certificateutil/info_model.go
```diff
@@ -22,32 +22,21 @@ type CertificateInfoModel struct {
...
-	printable["expire"] = info.EndDate.String()
-
-	errors := []string{}
-	if err := info.CheckValidity(); err != nil {
-		errors = append(errors, err.Error())
-	}
-	if len(errors) > 0 {
-		printable["errors"] = errors
-	}
+	team := fmt.Sprintf("%s (%s)", info.TeamName, info.TeamID)
```
lpusok: Check Cert and profile installer step logging

Gopkg.toml
```diff
@@ -0,0 +1,31 @@
+[[constraint]]
+  branch = "pkcs12"
```
godrei: do not forget to change back to master

certificateutil/filter.go
```diff
@@ -13,22 +13,40 @@ func FilterCertificateInfoModelsByFilterFunc(certificates []CertificateInfoModel
...
-				certificateInfosByName[certificateInfo.CommonName] = certificateInfo
-			}
+		if certificateInfo.CheckValidity() != nil {
+			invalidCertificates = append(invalidCertificates, certificateInfo)
+			continue
+		}
+		activeCertificate, ok := certificateInfosByName[certificateInfo.CommonName]
+		if !ok {
+			certificateInfosByName[certificateInfo.CommonName] = certificateInfo
+		} else if certificateInfo.EndDate.After(activeCertificate.EndDate) {
```
godrei: what if a cert with common name A expires 2016.01.02, found first, then
cert with common name A expires 2016.01.01 <- this will not be added to the `duplicatedCertificates`, since the expire is before the first cert's expire

certificateutil/info_model.go
```diff
@@ -112,7 +103,15 @@ func InstalledCodesigningCertificateInfos() ([]CertificateInfoModel, error) {
 	if err != nil {
 		return nil, err
 	}
-	return CertificateInfos(certificates), nil
+
+	infos := []CertificateInfoModel{}
```
godrei: why was this change needed?

certificateutil/info_model.go
```diff
@@ -80,30 +74,27 @@ func NewCertificateInfo(certificate x509.Certificate) CertificateInfoModel {
...
 // NewCertificateInfosFromPKCS12 ...
 func NewCertificateInfosFromPKCS12(pkcs12Pth, password string) ([]CertificateInfoModel, error) {
-	certificates, err := CertificatesFromPKCS12File(pkcs12Pth, password)
+	identities, err := CertificatesFromPKCS12File(pkcs12Pth, password)
 	if err != nil {
 		return nil, err
 	}
-	return CertificateInfos(certificates), nil
+
+	infos := []CertificateInfoModel{}
```
godrei: InstalledCodesigningCertificates

certificateutil/info_model.go
lpusok: certificate and profile installer does not use this function.

certificateutil/info_model.go
lpusok: removed NewCertificateInfosFromPKCS12 function

certificateutil/filter.go
lpusok: fixed

certificateutil/filter.go
```diff
@@ -22,26 +22,36 @@ type ValidCertificateInfo struct {
...
-	validCertificates := []CertificateInfoModel{}
-	for _, validCertificate := range certificateInfosByName {
-		validCertificates = append(validCertificates, validCertificate)
+	var validCertificates, duplicatedCertificates []CertificateInfoModel
+	for _, certs := range nameToCerts {
+		if len(certs) == 0 {
+			continue
+		}
+
+		latestCert := certs[0]
```
godrei: NOTE: what about this

```
		sort.Slice(certs, func(i, j int) bool {
			return certs[i].EndDate.Before(certs[j].EndDate)
		})

		validCertificates = append(validCertificates, certs[0])
		duplicatedCertificates = append(duplicatedCertificates, certs[1:]...)
```

is not it a bit cleaner?

lpusok: It is more clear, even if less optimal in theory.

Gopkg.toml
lpusok: fixed

PR #84, opened 2019-07-08
TOOL-975 #Updated dependencies.

richard-bogdan-bitrise:


xcarchive/xcarchive.go
```diff
@@ -39,3 +42,45 @@ func IsMacOS(archPath string) (bool, error) {
...
+// CheckForXcarchive checks if the given zip is an xcarhive or not
+func CheckForXcarchive(pth string) bool {
+	filename := filepath.Base(pth)
+	s := strings.Split(filename, ".")
+	length := len(s)
+	if length < 2 {
+		return false
+	}
+	xc := s[length-2]
+	if xc == "xcarchieve" {
```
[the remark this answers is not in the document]
trapacska: but instead of this block please use strings.HasSuffix, filepath.Ext, or any other extansion check built-in

ipa/ipa.go
```diff
@@ -1,47 +1,14 @@
...
 
 import (
-	"fmt"
 	"path/filepath"
 	"strings"
 
 	"github.com/bitrise-io/go-utils/pathutil"
 	"github.com/bitrise-io/go-utils/ziputil"
-	"github.com/bitrise-io/go-xcode/utility"
+	pth "github.com/bitrise-io/go-xcode/utility"
```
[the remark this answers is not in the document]
richard-bogdan-bitrise: Removed.

ipa/ipa_test.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/bitrise-io/go-utils/fileutil"
 	"github.com/bitrise-io/go-utils/pathutil"
+	pthUtil "github.com/bitrise-io/go-xcode/utility"
```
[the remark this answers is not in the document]
richard-bogdan-bitrise: Removed.

xcarchive/xcarchive.go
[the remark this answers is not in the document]
richard-bogdan-bitrise: Updated code.

xcarchive/xcarchive.go
```diff
@@ -2,10 +2,13 @@ package xcarchive
 
 import (
 	"path/filepath"
+	"strings"
 
 	"github.com/bitrise-io/go-utils/log"
 	"github.com/bitrise-io/go-utils/pathutil"
+	"github.com/bitrise-io/go-utils/ziputil"
 	"github.com/bitrise-io/go-xcode/plistutil"
+	pth "github.com/bitrise-io/go-xcode/utility"
```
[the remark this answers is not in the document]
richard-bogdan-bitrise: Fixed.

xcarchive/xcarchive.go
```diff
@@ -39,3 +42,45 @@ func IsMacOS(archPath string) (bool, error) {
 
 	return exist, nil
 }
+
+func unwrapFileEmbeddedInXcarchiveDir(xcarchivePth, fileName string) (string, error) {
+	// example path: xcarhive.zip/xcarhive/Products/Applications/sample.app/Contents/Info.plist
```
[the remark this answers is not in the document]
richard-bogdan-bitrise: Removed.

xcarchive/xcarchive.go
```diff
@@ -39,3 +42,38 @@ func IsMacOS(archPath string) (bool, error) {
 
 	return exist, nil
 }
+
+func unwrapFileEmbeddedInXcarchiveDir(xcarchivePth, fileName string) (string, error) {
```
trapacska: I think this is really not common to have only one file from an archive, and with this solution, each time you are looking for a file will create a new temp dir and unzip the archive. Try to find a solution where you can give an array of files in and then receive the files back somehow. So you can skip the extracting step each time.

xcarchive/xcarchive.go
```diff
@@ -39,3 +42,38 @@ func IsMacOS(archPath string) (bool, error) {
...
+
+	applicationsPth := filepath.Join(tmpDir, filepath.Base(xcarchivePth), "Products", "Applications")
+	appName := strings.TrimSuffix(filepath.Base(xcarchivePth), filepath.Ext(xcarchivePth))
+
+	return utility.FindFileInAppDir(applicationsPth, appName, fileName)
+}
+
+// UnwrapEmbeddedMobileProvision ...
+func UnwrapEmbeddedMobileProvision(xcarchivePth string) (string, error) {
+	return unwrapFileEmbeddedInXcarchiveDir(xcarchivePth, "embedded.mobileprovision")
```
trapacska: I think these wrappers are not really needed, what about having the unwrapFileEmbeddedInXcarchiveDir exported only?

xcarchive/xcarchive.go
trapacska: this method gets file paths that are unzipped, this is not a logic strictly for xcarchives, this matching for any zip related logic, please remove xcarchive labels from everywhere in this case

utility/path.go
```diff
@@ -67,3 +70,36 @@ func BaseFilter(base string, allowed bool) FilterFunc {
 		return (allowed == strings.EqualFold(base, b)), nil
 	}
 }
+
+// FindFileInAppDir ...
+func FindFileInAppDir(basePth, preferedAppName, fileName string) (string, error) {
```
trapacska: on osx an app extension is a dir simply, please merge the basepath and preferedAppName arguments as thos are going to be in the same path anyway, and expect a filepath with the suffix of .app

xcarchive/xcarchive.go
```diff
@@ -39,3 +42,38 @@ func IsMacOS(archPath string) (bool, error) {
...
+}
+
+// UnwrapEmbeddedInfoPlist ...
+func UnwrapEmbeddedInfoPlist(xcarchivePth string) (string, error) {
+	return unwrapFileEmbeddedInXcarchiveDir(xcarchivePth, "Info.plist")
+}
+
+// CheckForXcarchive checks if the given zip is an xcarhive or not
+func CheckForXcarchive(pth string) bool {
+	filename := filepath.Base(pth)
```
trapacska: you no need to get the base of a file path, because the extension is going to be the suffix anyway, independently of the length of the path

xcarchive/xcarchive.go
```diff
@@ -39,3 +42,38 @@ func IsMacOS(archPath string) (bool, error) {
...
+	if strings.HasSuffix(filename, ".xcarchive.zip") {
```
trapacska: here you can return with `return strings.HasSuffix(pth, ".xcarchive.zip")` but please do not keep this exported function for a single line of code

xcarchive/xcarchive.go
richard-bogdan-bitrise: Similarly to unwrapFileEmbeddedInPayloadAppDir in ipa.go, the purpose of this method is not to grant a universal solution for unwrapping, but to have a solution for xcarhives/ipas, where the needed files are in given folders. For reducing code duplication, there should be a general method, and these methods should use that. What do you think?

xcarchive/xcarchive.go
richard-bogdan-bitrise: Wrappers usually increase readability, reusability and are less error prone.

utility/path.go
richard-bogdan-bitrise: Updated

xcarchive/xcarchive.go
richard-bogdan-bitrise: Removed.

xcarchive/xcarchive.go
richard-bogdan-bitrise: In case of utilities, I prefer to use wrappers, as they are reusable, and less error prone, so I would keep this method. What do you think?

xcarchive/xcarchive.go
richard-bogdan-bitrise: Updated code, used different approach.

xcarchive/xcarchive.go
richard-bogdan-bitrise: As discussed, removed, and used directly in deploy step.

utility/path.go
```diff
@@ -67,3 +70,35 @@ func BaseFilter(base string, allowed bool) FilterFunc {
 		return (allowed == strings.EqualFold(base, b)), nil
 	}
 }
+
+// FindFileInAppDir ...
+func FindFileInAppDir(appDir, fileName string) (string, error) {
+	filePth := filepath.Join(appDir+".app", fileName)
```
trapacska: appDir should be a path that already contains the .app extension

xcarchive/xcarchive.go
```diff
@@ -39,3 +41,30 @@ func IsMacOS(archPath string) (bool, error) {
...
 }
+
+// UnzipXcarchive ...
+func UnzipXcarchive(xcarchivePth string) (string, error) {
+	tmpDir, err := pathutil.NormalizedOSTempDirPath("__xcarhive__")
+	if err != nil {
+		return "", err
+	}
+
+	if err := ziputil.UnZip(xcarchivePth, tmpDir); err != nil {
```
trapacska: please move to the end of the block, instead of returning nil just
return tmpDir, ziputil.UnZip(xcarchivePth, tmpDir)

utility/path.go
richard-bogdan-bitrise: Fixed.

xcarchive/xcarchive.go
richard-bogdan-bitrise: Fixed.

PR #86, opened 2019-10-08
TOOL-368 MacOS XcArchive embedded profile location fix

BirmacherAkos:
- The package searched on a wrong location for a wrong provisioning profile extension
`Contents/Resources/embedded.mobileprovision` => `Contents/embedded.provisionprofile`

### DIFF
--- a/xcarchive/macos.go
+++ b/xcarchive/macos.go
@@ -43,7 +43,7 @@ func newMacosBaseApplication(path string) (macosBaseApplication, error) {
 
 	var provisioningProfile *profileutil.ProvisioningProfileInfoModel
 	{
-		provisioningProfilePath := filepath.Join(path, "Contents/Resources/embedded.mobileprovision")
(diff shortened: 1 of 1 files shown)


xcarchive/macos.go
```diff
@@ -43,7 +43,7 @@ func newMacosBaseApplication(path string) (macosBaseApplication, error) {
 
 	var provisioningProfile *profileutil.ProvisioningProfileInfoModel
 	{
-		provisioningProfilePath := filepath.Join(path, "Contents/Resources/embedded.mobileprovision")
+		provisioningProfilePath := filepath.Join(path, "Contents/embedded.provisionprofile")
```
trapacska: Is it for all versions?

BirmacherAkos: About the extension I'm sure, but about the location... 🤔 
I have to check it with Xcode 9 too, just to be sure 😭

PR #96, opened 2020-12-15
Expose HasAppClip on Targets

adborbas:
## Context
Preparing for the support of AppClips in the ProjectScanner.

Ticket: [STEP-87](https://bitrise.atlassian.net/browse/STEP-87)

## Files to check
- Do **NOT** check `vendor/` directory

## Changes
- Extended `SchemeModel` with `BuildableReferenceIDs` and `TargetModel` with `ID` to provide a way to map targets to schemes.
- Extended `TargetModel` with `HasAppClip` which shows whether a target has an AppClip target as its dependency.
- Added tests for retrieving targets from an `.xcodeproj`

## Investigation details
- Added `github.com/bitrise-io/xcode-project/xcscheme` as a dependency to parse the `BuildableReferenceIDs` in the xcscheme. This should be a temporarily workaround until we are depending in xcode-proj in the `ProjectScanner`.


xcodeproj/xctarget_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/bitrise-io/go-utils/fileutil"
+	"github.com/bitrise-io/go-utils/pathutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_GivenOnlyApp_WhenProjectTargetCalled_ThenExpectSingleTarget(t *testing.T) {
```
lpusok, line 14: NOTE: This test naming is quite new, in other places is usually just Test_ProjectTarget() fro example. It could also be a table driven testm where the name of the test could contain the same infos (GivenOnlyApp_WhenProjectTargetCalled_ThenExpectSingleTarget)

lpusok (approved):

xcodeproj/xctarget_test.go
adborbas: Fair point and I have wanted to do it that way but I've felt that the `expectedTargets` are too big to fit into that scheme and this way would be more readable.

PR #97, opened 2021-01-04
Generate sample projects

adborbas:
## Context

Use [tuist](https://tuist.io/) to generate the sample iOS apps for testing instead of keeping `.pbxproj` contents in the repo.

Ticket: [STEP-265](https://bitrise.atlassian.net/browse/STEP-265)

## Changes

- Updated `xctarget_test`s to use generated projects.
- Placed project template descriptions into `xcodeproj/_test/template`
- Version `tuist` with `.tuist-version` file.
- Added `resolve-tuist` workflow to bitrise.yml to install Tuist on the CI and install/update Tuist on the local machine.

## Investigation details
Please check the referenced Confluence page in the ticket.


tuistupdater.sh
```diff
@@ -0,0 +1,34 @@
...
+# `tuist update` exists with an error code although
+# it actually succeeds.
+set -e
+
+echo ""
+echo "Check if local Tuist needs to be updated"
+GLOBAL_TUIST=$(cd / && tuist version)
+echo "Global tuist version: $GLOBAL_TUIST"
+
+LOCAL_TUIST=$(cat .tuist-version)
```
mateherber, line 23: Why do we store the .tuist-version and push it to the repository if we update tuist every time anyway.

mateherber: This approach is really sick imho, great job 👍

mateherber (approved):

tuistupdater.sh
adborbas: `.tuist-version` makes sure that we are using the same version of tuist locally and on the CI. To keep up with the latest tuist version, before running the test we check for any updates locally. This way we do not have to manually check for tuist updates.

mateherber: ic, thx

PR #98, opened 2021-01-11
Returning error upon missing dSYM removed

istvankovacs-bitrise:
https://bitrise.atlassian.net/browse/STEP-303

### Context
Based on [this](https://developer.apple.com/documentation/xcode/building_your_app_to_include_debugging_information#3403353) document the generation of the build symbols is optional, so there is no need to return an error if there is not any dSYM found.


adborbas (approved):
👍

PR #105, opened 2021-05-18
Step 821 migrate packages

mateherber:


istvankovacs-bitrise (approved):

utility/path.go
```diff
@@ -4,38 +4,12 @@ import (
...
-			filtered = append(filtered, pth)
-		}
-	}
-
-	return filtered, nil
-}
-
 // ListEntries ...
-func ListEntries(dir string, filters ...FilterFunc) ([]string, error) {
+func ListEntries(dir string, filters ...pathutil.FilterFunc) ([]string, error) {
```
godrei, line 12: Should not it go to `go-utils/pathutil`? I feel like this function is generic enough and has no ios specific logic.

godrei: The same applies to `EscapeGlobPath`, I suggest moving it to go-utils/pathutil

lpusok (dismissed):

lpusok (approved):

PR #107, opened 2021-06-02
Step 917 xcode project update

godrei:
This PR adds support for iOS Auto Provision step's UITest target signing, see: https://github.com/bitrise-steplib/steps-ios-auto-provision-appstoreconnect/pull/51

Changes:
- removed invalid `includeUITest` input from `Target.DependentExecutableProductTargets` (the executable target has no dependency on the related UITest targets, it works the other way around)
- new `DependesOn` returns if the target depends on another target
- `InformationPropertyList` renamed to `Infoplist` based on the `INFOPLIST_FILE` env

### DIFF
--- a/xcodeproject/xcodeproj/target.go
+++ b/xcodeproject/xcodeproj/target.go
@@ -44,17 +44,17 @@ func (t Target) DependentTargets() []Target {
 }
 
 // DependentExecutableProductTargets ...
-func (t Target) DependentExecutableProductTargets(includeUITest bool) []Target {
+func (t Target) DependentExecutableProductTargets() []Target {
 	var targets []Target
 	for _, targetDependency := range t.Dependencies {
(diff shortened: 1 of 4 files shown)


xcodeproject/xcodeproj/target.go
```diff
@@ -43,18 +43,29 @@ func (t Target) DependentTargets() []Target {
 	return targets
 }
 
+// DependesOn ...
+func (t Target) DependesOn(targetID string) bool {
```
lpusok, line 47: Do we care about indirect dependencies (2 level down?)

godrei: For now we do not.
iOS Auto Prov step only signs the UITest targets depending on the main application target.

lpusok (approved):

PR #109, opened 2021-07-07
Add Recreate Shared Schemes feature

lpusok:
### Context

Add support to write Schemes and the feature to recreate Schemes from Targets.

### Changes

- Add Marshal method to Scheme: This Marshals a usual XML file but modifies it to match the format used by Xcode
- Support all fields for Scheme (including LaunchAction, ProfileAction and ArchiveAction)
- Adding ReCreateSchemes function

### DIFF
--- a/xcodeproject/xcscheme/xcscheme.go
+++ b/xcodeproject/xcscheme/xcscheme.go
@@ -3,6 +3,7 @@ package xcscheme
 import (
 	"encoding/xml"
 	"fmt"
+	"io/ioutil"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


xcodeproject/xcscheme/xcscheme.go
```diff
@@ -92,6 +162,58 @@ func Open(pth string) (Scheme, error) {
...
+)
+
+// Marshal ...
+func (s Scheme) Marshal() ([]byte, error) {
+	contents, err := xml.Marshal(s)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal Scheme: %v", err)
+	}
+
+	contentsNewline := strings.ReplaceAll(string(contents), "><", ">\n<")
```
mateherber, line 185: What about "> <" or there is no such thing?

mateherber (approved):

xcodeproject/xcscheme/xcscheme.go
lpusok: It could be valid, but it is not generated by the go xml package (in the unformatted output).

PR #111, opened 2021-07-22
Add workaround for simulator boot failure

benbitrise:
Big Sur sometimes fails to open the Simulator App. This workaround reregisters the app in launch services database.

The workaround only runs on Big Sur to limit the blast radius of a future change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 _tmp/
-.vscode/*
+.vscode/
+/.idea/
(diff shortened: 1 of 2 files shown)


ofalvai:
Closing this PR in favor of https://github.com/bitrise-steplib/steps-xcode-test/pull/177

This lib is not called in every code path, so it's better to include the workaround in the step itself.

PR #112, opened 2021-08-30
Migrate to the new go-utils and go-steputils packages

godrei:
### Context

This PR migrates go-xcode to the new go-utils and go-steputils packages in order to be able to pull these packages in Xcode Test step.


lpusok (dismissed):
Do you plan to add commandFactory as additional function parameter?

godrei:
@lpusok it is not yet clear to me in all cases if we should pass commandFactory as a function parameter, or we should add the functions to a struct and pass as a struct property.

godrei:
@lpusok  in go-steputils we have a global non-exported variable holding the default factory: https://github.com/bitrise-io/go-steputils/blob/master/command/rubycommand/rubycommand.go#L39

this allows white-box testing. Do you think we should do the same here?

lpusok:
@godrei Do not a have a strong opinion on adding the the global variable, I am more curios on wether the xcode-test Step tests require it. Personally I prefer if we would change the functions to not require the Command factory at all, where possible. But I agree that that is not possible for every function.

godrei:
@lpusok `utility.GetXcodeVersion` receives command.Factory as a function parameter, as this is used in the xcode-test step.
Other packages got a temporary global variable storing a command.Factory instance.

lpusok (approved):

PR #129, opened 2021-11-09
[MOB-140] Cache and xcpretty refactor

tothszabi:
This PR contains two changes:

1.) `SwiftPackageCache`

This thin wrapper was used in the xcode-test step because the main swift cache related methods are single functions. I though that any step which wants to write tests will need the same thing so I just moved it into this repo. I deprecated the global functions but also updated then to use the struct instead.

2.) `Xcpretty`

Did a similar thing as above. Xcpretty had to be mocked for the unit tests so I had to group the global functions in a struct.


xcodecache/swiftpm_cache.go
```diff
@@ -51,3 +65,13 @@ func CollectSwiftPackages(xcodeProjectPath string) error {
 	}
 	return nil
 }
+
+// SwiftPackagesPath is deprecated. Please use the SwiftPackageCache interface instead.
```
ofalvai, line 69: Let's use the official deprecation syntax so that IDEs pick this up and highlight method usages: https://stackoverflow.com/questions/7849663/how-do-you-mark-code-as-deprecated-in-go/36360323

tothszabi: @ofalvai I temporary forgot why I have not used that. Pushed a new commit and now I know. 😄

So I was initially using that format but our linting rules permit it. This is the result:
```
/Users/vagrant/go/src/github.com/bitrise-io/go-xcode/xcodecache/swiftpm_cache.go:69:1: comment on exported function SwiftPackagesPath should be of the form "SwiftPackagesPath ..."
/Users/vagrant/go/src/github.com/bitrise-io/go-xcode/xcodecache/swiftpm_cache.go:74:1: comment on exported function CollectSwiftPackages should be of the form "CollectSwiftPackages ..."
Found 2 lint suggestions; failing.
golint failed: exit status 1
```

Should we change the linting our do what I did originally with the comments?

ofalvai: I would follow this pattern in the short term: https://github.com/bitrise-io/go-utils/blob/master/log/log.go#L159

But we definitely need to get rid of this rule :)

tothszabi: That is a good tip. It did not occur to me to split the comment that way. Thanks.

ofalvai (approved):

PR #130, opened 2021-11-10
Read archive codesign parameters

godrei:
### Context

We are about to introduce automatic codesigning to xcodebuild related steps.
The current autocodesign package offers functionality to read the codesigning requirements from an iOS project.
In the case of the `Export iOS and tvOS Xcode archive` an iOS project is not given, but a `.xcarchive` file.

This PR adds a new function to `xcarchive` package (IosArchive.ReadCodesignParameters) to read codesigning requirements from an archive file. The return value can be passed to the autocodesign lib.

The archive file does not have any information about the project's UITest targets, so the `UITestTargetBundleIDs` property of the `autocodesign.AppLayout` return value will be nil always. This means `Export iOS and tvOS Xcode archive` step can not automatically codesign the project's UITest targets.

### Changes

- Add `IosArchive.ReadCodesignParameters` function to the `xcarchive` package.

### DIFF
--- a/xcarchive/ios.go
+++ b/xcarchive/ios.go
@@ -380,6 +380,7 @@ func (archive IosArchive) FindDSYMs() ([]string, []string, error) {
 	return findDSYMs(archive.Path)
 }
 
+// ReadCodesignParameters ...
 func (archive IosArchive) ReadCodesignParameters() (*autocodesign.AppLayout, error) {
 	var teamID string
 	entitlementsMap := map[string]autocodesign.Entitlements{}
--- a/xcarchive/ios.go
+++ b/xcarchive/ios.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 
 	"github.com/bitrise-io/go-utils/pathutil"
+	"github.com/bitrise-io/go-xcode/autocodesign"
 	"github.com/bitrise-io/go-xcode/plistutil"
 	"github.com/bitrise-io/go-xcode/profileutil"
 )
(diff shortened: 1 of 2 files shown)


xcarchive/ios.go
```diff
@@ -378,3 +379,38 @@ func (archive IosArchive) BundleIDProfileInfoMap() map[string]profileutil.Provis
...
+	var teamID string
+	entitlementsMap := map[string]autocodesign.Entitlements{}
+
+	bundleIDProfileInfoMap := archive.BundleIDProfileInfoMap()
+	for bundleIdentifier, provisioningProfile := range bundleIDProfileInfoMap {
+		if teamID == "" {
+			teamID = provisioningProfile.TeamID
+		}
+
+		entitlementsMap[bundleIdentifier] = autocodesign.Entitlements(provisioningProfile.Entitlements)
```
godrei, line 394: the Provisioning Profile's entitlements contains not just codesign-related entitlements (like `get-task-allow`), we should either filter them here or later when using the entitlement.

tothszabi: Good point. Had a look at the other usage of the `AppLayout` struct and (if I read it correctly) it returns only the user specified entitlements from the actual project itself. Based on that I would replicate the same behaviour here and only return the codesign related ones.

xcarchive/ios.go
```diff
@@ -378,3 +379,38 @@ func (archive IosArchive) BundleIDProfileInfoMap() map[string]profileutil.Provis
...
+	}
+
+	var platform autocodesign.Platform
+
+	platformName := archive.Application.InfoPlist["DTPlatformName"]
```
lpusok, line 399: Can maybe this and https://github.com/bitrise-steplib/steps-deploy-to-itunesconnect-application-loader/blob/245a973d4c4655ddf37ff491c3e52d17ebec6259/main.go#L76 be extracted to the same helper function?

xcarchive/ios.go
```diff
@@ -378,3 +379,38 @@ func (archive IosArchive) BundleIDProfileInfoMap() map[string]profileutil.Provis
 func (archive IosArchive) FindDSYMs() ([]string, []string, error) {
 	return findDSYMs(archive.Path)
 }
+
+// ReadCodesignParameters ...
```
lpusok, line 383: Is there explicit support needed for App Clips?

xcarchive/ios.go
godrei: What do you think, what and how should we share here?

As I see, only this tiny part could be shared: https://github.com/bitrise-steplib/steps-deploy-to-itunesconnect-application-loader/blob/245a973d4c4655ddf37ff491c3e52d17ebec6259/main.go#L91-L107

But, in the case of an xcarchive `*simulator` platforms are not valid. Also `watchos` does not seem to be a valid platform if you read an archive file (even if you are building a standalone watchos application the platform name is: `iphoneos`)

xcarchive/ios.go
godrei: I do not think so, you can not create a valid archive containing an AppClip as the main Product. 
Actually, you can create an AppClip archive but it seems to be invalid (it does not contain any Product or executable file, probably an Xcode bug).

xcarchive/ios.go
godrei: discussed offline that we will leave it as it is currently.

xcarchive/ios.go
lpusok: Thanks for checking, based on this you can leave as is.

xcarchive/ios.go
lpusok: Okay, good to know.

lpusok (approved):

PR #132, opened 2021-11-15
Support Xcode Code Signing

lpusok:
Changes required for supporting authenticated xcodebuild command. Contains detection of the project is Xcode managed signing (to decide between Bitrise and Xcode Code Signing method).
Exported device registration functions, so those are available even with Xcode Code Signing.

Issue: https://linear.app/bitrise/issue/MOB-229/xcode-archive-step-introduce-xcodebuild-automatic-code-signing

## Changes
- Added authentication params support for xcodebuild command
- Added projecthelper.IsSigningManagedAutomatically(), to decide between Bitrise and Xcode Code Signing method
- Exported device registration functions, so those are available even with Xcode Code Signing.


xcodebuild/xcodebuild.go
```diff
@@ -11,3 +11,19 @@ type CommandModel interface {
 	PrintableCmd() string
 	Command(opts *command.Opts) command.Command
 }
+
+// AuthenticationParams ...
```
ofalvai, line 15:
```suggestion
// AuthenticationParams are used to authenticate to App Store Connect API and let xcodebuild download missing provisioning profiles.
```

ofalvai (approved):

PR #133, opened 2021-11-19
Retry profile creation

lpusok:
Try to detect if a profile was deleted by a parallel actor. This kinds of race condition can happen as the underlying Developer Portal account is the same.

Introduced *ProfilesInconsistentError* that can be returned by both _spaceship_ and _appstoreconnect_.
The retry mechanism is in **ensureProfile**, so the profile is looked up again. In case it was deleted, retrying just the failing call (to download it or fetch Device IDs) would never succeed.

An example of it working is in https://app.bitrise.io/build/2881bc6e-e849-461e-bcb0-3054ef8fa3b2#?tab=log.


shams-ahmed (approved):

autocodesign/devportalclient/spaceship/profiles.go
```diff
@@ -174,34 +174,43 @@ func (c *ProfileClient) DeleteProfile(id string) error {
...
-		profileTypeArgKey, string(profileType),
-	)
-	if err != nil {
-		return nil, err
-	}
-
-	output, err := runSpaceshipCommand(cmd)
-	if err != nil {
-		return nil, err
+	for i := 0; i < 6; i++ {
```
tothszabi, line 177: I probably do not know all of the details of this workaround and that is why I will be asking asking these questions. 

Why is this 6? It seems like an arbitrary high number to me. I mean in my mind I was comparing this to a scenario where the server returns a 500 error and we will still make 5 request just to be sure it is down. I know it is not a good example but could not come up with a better one. So retrying only once is not enough? Or even better did you find out why spaceship can end up returning an empty profile?

Another thing is that could this for loop be converted to some retry functionality we have in the util repos? I am not fully aware what feature set the current retry mechanism has so just asking.

lpusok: Did not find out the root cause :/. One retry is definitely not enough, seen at least 3 retry needed (https://app.bitrise.io/build/2e08da10-5813-475c-a50c-cf3164be6409#?tab=log). 5 could work, was just not confident that it is enough. The retry util suggestion makes sense, as it would wait for a few seconds (which may help).

tothszabi: Knowing these I am happy to have a high number there. I might even say to round it up to 10 to make sure we catch all these cases. 

I prefer to have comments in code for such code pieces (the ones which is not clear at first look). So I would even suggest to add a comment above the retry which is something like:
```
// This retry is needed because spaceship can return an empty profile. In local testing at least 3 retries were needed in some cases. The {{insert final high retry number here}} retries probably will be never needed but only act as a big buffer to mitigate all occurrences.
```
I am not keen on the wording so feel free to rephrase it. 😄

tothszabi (approved):

autocodesign/devportalclient/appstoreconnectclient/profiles.go
```diff
@@ -342,11 +342,9 @@ func (c *ProfileClient) SyncBundleID(bundleID appstoreconnect.BundleID, appEntit
 }
 
 func wrapInProfileError(err error) error {
-	if respErr, ok := err.(appstoreconnect.ErrorResponse); ok {
```
godrei, line 345: NOTE: should not we use errors.Is or errors.As instead of the type cast?

autocodesign/devportalclient/spaceship/profiles.go
```diff
@@ -196,6 +197,10 @@ func (c *ProfileClient) CreateProfile(name string, profileType appstoreconnect.P
 		return nil, fmt.Errorf("failed to unmarshal response: %v (%s)", err, output)
 	}
 
+	if profileResponse.Data.Name == "" {
+		return nil, autocodesign.NewProfilesInconsistentError(errors.New("empty profile generated by Spaceship"))
```
godrei, line 201: isn't just the name empty? i would remove the `by Spaceship` suffix.

autocodesign/devportalclient/appstoreconnectclient/profiles.go
lpusok: updated

autocodesign/profiles.go
```diff
@@ -199,14 +201,43 @@ func (m profileManager) ensureBundleID(bundleIDIdentifier string, entitlements E
...
 	m.bundleIDByBundleIDIdentifer[bundleIDIdentifier] = bundleID
 
 	return bundleID, nil
 }
 
+func (m profileManager) ensureProfileWithRetry(profileType appstoreconnect.ProfileType, bundleIDIdentifier string, entitlements Entitlements, certIDs, deviceIDs []string, minProfileDaysValid int) (*Profile, error) {
+	var profile *Profile
+	// Accessing the same Apple Developer Portal team can cause race conditions (parallel CI runs for example).
+	// Between the time of finding and downloading a profile, it could have been deleted for example.
+	if err := retry.Times(5).Wait(10 * time.Second).TryWithAbort(func(attempt uint) (error, bool) {
```
tothszabi, line 216: Should we maybe switch up the numbers here?

Right now it does a low amount of retries with a large waiting period between them. If we do `Times(10).Wait(5 * time.Second)` then we could waiting times at least in half which means we could gain a few seconds for the user. 

You mentioned earlier that you saw that on some occasions on the profile was not empty on the third try and also wrote in the comment that this also comes up when there are multiple parallel accesses to the portal. So I am not sure in how long tasks take on the portal but let me throw in the following idea and hear your opinion about it.

What if we do 
```
if err := retry.Times(10).Wait(3 * time.Second).TryWithAbort(func(attempt uint) (error, bool) {
```

autocodesign/devportalclient/appstoreconnectclient/profiles.go
tothszabi: Just a question about this piece of code because you wrote `updated` but the type casting is still there. Is it on purpose?

godrei (dismissed):

autocodesign/profiles.go
lpusok: I would go with 10 seconds as the retry only happens if there is someone else editing the Developer Portal. So waiting a bit should help so they can finish whatever they are doing. I would consider an exponential backoff (3 -> 6 -> 12) also.

autocodesign/devportalclient/appstoreconnectclient/profiles.go
lpusok: It was a mistake, my bad.

tothszabi (approved):

PR #134, opened 2021-11-22
Xcode Code Signing support

lpusok:
- Adding function to write private key to file
- Adding codesign package
- Moved codesign logic


autocodesign/example_test.go
```diff
@@ -1,96 +0,0 @@
-package autocodesign_test
```
godrei, line 1: why did we remove the example?

autocodesign/projectmanager/projectmanager.go
```diff
@@ -17,13 +17,28 @@ type Project struct {
...
+
 // InitParams ...
 type InitParams struct {
 	ProjectOrWorkspacePath string
 	SchemeName             string
 	ConfigurationName      string
 }
 
+// NewFactory ...
+func NewFactory(params InitParams) Factory {
```
godrei, line 33: why is this factory needed?

codesign/codesign.go
```diff
@@ -0,0 +1,338 @@
...
+			}
+
+			return appleauth.Credentials{}, fmt.Errorf("API key authentication is selected in Step inputs, but Bitrise Apple service connection is unset")
+		}
+
+		return appleauth.Credentials{}, fmt.Errorf("could not select Apple authentication credentials: %w", err)
+	}
+
+	if authConfig.APIKey != nil {
+		authConfig.AppleID = nil
```
godrei, line 207: for the sake of clarity should not we clear `authConfig.APIKey` if `authConfig.AppleID != nil` ?

codesign/codesign.go
```diff
@@ -0,0 +1,338 @@
...
+	certificates, err := m.certDownloader.GetCertificates()
+	if err != nil {
+		return fmt.Errorf("failed to download certificates: %s", err)
+	}
+
+	m.logger.Infof("Installing downloaded certificates:")
+	for _, cert := range certificates {
+		m.logger.Printf("- %s", cert)
+		// Empty passphrase provided, as already parsed certificate + private key
+		if err := m.keychain.InstallCertificate(cert, ""); err != nil {
```
godrei, line 265: in case of Bitrise managed auto codesigning we install a single certificate per distribution type, should not we do so here as well?

codesign/inputparse.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		keychainP, err := keychain.New(p.KeychainPath, p.KeychainPassword, p.CommandFactory)
+		if err != nil {
+			return ParsedConfig{}, fmt.Errorf("failed to open keychain: %s", err)
+		}
+		keychainWriter = *keychainP
+	}
+
+	return ParsedConfig{
```
godrei, line 65: is ParsedConfig valid without a CertificatesAndPassphrases and Keychain (in case of authType == NoAuth)?

devportalservice/devportalservice.go
```diff
@@ -254,3 +255,21 @@ func validateTestDevice(deviceList []TestDevice) (validDevices, duplicatedDevice
 
 	return validDevices, duplicatedDevices
 }
+
+// WritePrivateKeyToFile writes the contents of the private key to a temporary file and returns its path
+func (c *APIKeyConnection) WritePrivateKeyToFile() (string, error) {
+	privatekeyFile, err := os.CreateTemp("", "apiKey*.p8")
```
godrei, line 261: why is the base name `apiKey*.p8` it is usually in `AuthKey_<ISSUER_ID>.p8` format?

codesign/codesign.go
lpusok: Would not select a single one as Xcode can select one later, but at least validate that development cert is avaialble (if needed)

codesign/inputparse.go
```diff
@@ -0,0 +1,111 @@
...
+	CertificateURLList        string
+	CertificatePassphraseList stepconf.Secret
+	KeychainPath              string
+	KeychainPassword          stepconf.Secret
+
+	CommandFactory command.Factory
+}
+
+// ParsedConfig ...
+type ParsedConfig struct {
```
godrei, line 28: Do we need the `Parsed` prefix?

codesign/inputparse.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/bitrise-io/go-steputils/stepconf"
+	"github.com/bitrise-io/go-utils/command"
+	"github.com/bitrise-io/go-utils/sliceutil"
+	"github.com/bitrise-io/go-xcode/autocodesign"
+	"github.com/bitrise-io/go-xcode/autocodesign/certdownloader"
+	"github.com/bitrise-io/go-xcode/autocodesign/keychain"
+)
+
+// StepInputParser ...
+type StepInputParser struct {
```
godrei, line 16: Why does StepInputParser store the inputs, while StepInputParser.Parse does not have any parameter?

I would move code signing related Step Inputs into a separate struct (codesign.Inputs) and call the Parse return value as codesign.Config

codesign/codesign.go
```diff
@@ -0,0 +1,338 @@
...
+
+	RegisterTestDevices bool
+	SignUITests         bool
+	MinProfileValidity  int
+	IsVerboseLog        bool
+}
+
+// Result ...
+type Result struct {
+	XcodebuildAuthParams *devportalservice.APIKeyConnection
```
godrei, line 55: Why is it wrapped into a new struct?

autocodesign/projectmanager/projectmanager.go
lpusok: Used this as a tool to allow for passing in the project unopened; parsing the project only when needed.

codesign/codesign.go
lpusok: For consistecy yes, should not really matter either way.

devportalservice/devportalservice.go
lpusok: This way an unique temp file, in case multiple Steps would try to create one.

codesign/codesign.go
lpusok: The  plan is to add other parameter that the standalone autoprov Step can use.

codesign/codesign_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+func newMockProject(isAutoSign bool, mockErr error) Project {
+	mockProjectHelper := new(mocks.Project)
+	mockProjectHelper.On("IsSigningManagedAutomatically", mock.Anything).Return(isAutoSign, mockErr)
+
+	return mockProjectHelper
+}
+
+func TestManager_downloadAndInstallCertificates(t *testing.T) {
+	// const teamID = "MYTEAMID"
```
godrei, line 104: do we need this comments?

lpusok: Fixed

autocodesign/example_test.go
lpusok: restored

godrei (approved):

PR #138, opened 2021-11-29
[MOB-112] Read entitlements from executable

tothszabi:
On iOS the `xcent` file is not included in the archive anymore. We need to have the required codesign entitlements ready for the handling the local codesign assets. This PR modifies the entitlement fetching logic to read it from the executable when the `xcent` file is missing.

It takes the the executable name from the `Info.plist` file because of the following edge case:
- The application could be renamed (I do not see why anyone would do that) and in that case the application bundle and the executable will have different names.
- On MacOS multiple executables could be packaged together with the main one.

If the above edge case happens then only the info file can help to discover the main binary name.


shams-ahmed (dismissed):
Looks good to me.. might be worth requesting a second review from someone experienced in code sign logic.

xcarchive/entitlements.go
```diff
@@ -0,0 +1,77 @@
...
+
+func executableNameFromInfoPlist(infoPlist plistutil.PlistData) string {
+	if name, ok := infoPlist["CFBundleExecutable"].(string); ok {
+		return name
+	}
+	return ""
+}
+
+func getEntitlements(basePath, entitlementsRelativePath, executableRelativePath string) (plistutil.PlistData, error) {
+	entitlements, err := entitlementsFromXcentFile(basePath, entitlementsRelativePath)
```
lpusok, line 21: Do we still need the xcent file logic at all?

xcarchive/entitlements.go
```diff
@@ -0,0 +1,77 @@
...
+	"path/filepath"
+
+	"github.com/bitrise-io/go-utils/command"
+	"github.com/bitrise-io/go-utils/env"
+	"github.com/bitrise-io/go-utils/pathutil"
+	"github.com/bitrise-io/go-xcode/plistutil"
+)
+
+func executableNameFromInfoPlist(infoPlist plistutil.PlistData) string {
+	if name, ok := infoPlist["CFBundleExecutable"].(string); ok {
```
godrei, line 14: NOTE: `infoPlist.GetString()` could be used here

xcarchive/entitlements.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return plistutil.PlistData{}, err
+	}
+
+	if entitlements != nil {
+		return *entitlements, nil
+	}
+
+	entitlements, err = entitlementsFromExecutable(basePath, executableRelativePath)
+
```
godrei, line 32: NOTE: error handling should follow the call without a newline

xcarchive/entitlements.go
```diff
@@ -0,0 +1,77 @@
...
+func entitlementsFromExecutable(basePath, executableRelativePath string) (*plistutil.PlistData, error) {
+	factory := command.NewFactory(env.NewRepository())
+	cmd := factory.Create("codesign", []string{"--display", "--entitlements", ":-", filepath.Join(basePath, executableRelativePath)}, nil)
+	entitlementsString, err := cmd.RunAndReturnTrimmedOutput()
+	if err != nil {
+		return nil, err
+	}
+
+	plist, err := plistutil.NewPlistDataFromContent(entitlementsString)
```
godrei, line 71: On my machine the first line of the output is: `Executable=<bin_path>`, is not it causing issues at plist parsing?

xcarchive/entitlements.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+func executableNameFromInfoPlist(infoPlist plistutil.PlistData) string {
+	if name, ok := infoPlist["CFBundleExecutable"].(string); ok {
+		return name
+	}
+	return ""
+}
+
+func getEntitlements(basePath, entitlementsRelativePath, executableRelativePath string) (plistutil.PlistData, error) {
```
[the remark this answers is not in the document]
tothszabi: In this case this function does not provide anything usable in itself. I think it is fine if it is just part of the 
```
func NewIosArchive(path string) (IosArchive, error)
```
public function call.

xcarchive/entitlements.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &plist, nil
+}
+
+func entitlementsFromExecutable(basePath, executableRelativePath string) (*plistutil.PlistData, error) {
+	factory := command.NewFactory(env.NewRepository())
+	cmd := factory.Create("codesign", []string{"--display", "--entitlements", ":-", filepath.Join(basePath, executableRelativePath)}, nil)
```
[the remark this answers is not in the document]
tothszabi: Will the `embedded.mobileprovision` have exactly the same entitlements in at? I thought the provisioning profile can have more entitlements than what the app actually is requiring. 

Apple still generates `xcent` files just does not include them in the archive. What it does now is to embed the contents of the xcent file into the binary with 
```
/usr/bin/codesign --force --sign FB4B4D09FF527AE3A2EA8D800C587E8BC58EC51D --entitlements /Users/szabi/Library/Developer/Xcode/DerivedData/Fruta-gnzpqmhzithwtvcqklttcjeydybd/Build/Intermediates.noindex/ArchiveIntermediates/Fruta\ iOS/IntermediateBuildFilesPath/Fruta.build/Release-iphoneos/Fruta\ iOS.build/Fruta\ iOS.app.xcent --generate-entitlement-der /Users/szabi/Library/Developer/Xcode/DerivedData/Fruta-gnzpqmhzithwtvcqklttcjeydybd/Build/Intermediates.noindex/ArchiveIntermediates/Fruta\ iOS/InstallationBuildProductsLocation/Applications/Fruta\ iOS.ap
```
so reading it from there is the same as if we would read it from the xcent file.

xcarchive/entitlements.go
tothszabi: I did not even notice it until you called it out. The tests work and they compare a real entitlement map read from the executable. 

I checked the command output and doing it this way
```
factory := command.NewFactory(env.NewRepository())
cmd := factory.Create("codesign", []string{"--display", "--entitlements", ":-", filepath.Join(basePath, executableRelativePath)}, nil)
entitlementsString, err := cmd.RunAndReturnTrimmedOutput()
```
only returns the actual xml entitlements string. The `Executable=<bin_path>` is missing. I do not see yet why this returns different output compared to executing it manually in the terminal. Do you have any idea?

xcarchive/entitlements.go
tothszabi: I kept it because in our artefacts repo the MacOS archive still has it. I will double check it.

xcarchive/entitlements.go
lpusok: Not sure if it is the case, but the codesign binary could check if it is running in an interactive terminal or not. I have seen brew CLI do that.

xcarchive/entitlements.go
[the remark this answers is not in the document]
lpusok: Yes the provisioning profile can have more entitlements indeed (tried out with a manually managed profile).

xcarchive/entitlements.go
tothszabi: Okay, too many open questions. I decided to open Hopper and have a look at the codesign internals. 

It seems like codesign can output various information like
- "Executable=%s"
- "Identifier=%s"
- "CodeDirectory v=%x size=%d flags=%s hashes=%d+%d location=%s"
- and so on

Every log seems like to have a log level. For the executable line it is 0 and something greater for all of the rest. That could be the reason why only the executable path is printed out.

The more interesting thing is that it uses the stderr to print out these informations. I think it is probably a workaround to separate the entitlements output from the verbose log messages because these logs probably should not go to the err output in normal circumstances. 

And this explains why `RunAndReturnTrimmedOutput` does not return it. I tried it with `RunAndReturnTrimmedCombinedOutput` and then the executable path was returned.

So I think we are good with this.

lpusok (dismissed):

xcarchive/entitlements.go
godrei: I checked it and with the latest Xcode version (13 at the moment), macOS archives behave the same way, xcent file is missing from the archive.

I'm going to remove the xcent file logic.

lpusok (approved):

lpusok (approved):

PR #141, opened 2021-12-02
Mob 112 archive codesign manager

godrei:
### Context

This PR introduces `codesign.Archive` structs, which implements `codesign.Project` interface, so that it can be used as codesign asset requirement source for `codesign.Manager`.

### Changes

- introduce `codesign.Archive` structs, which implements `codesign.Project` interface
- `codesign.Manager`'s Xcode project based constructor renamed to `NewManagerWithProject`
- `codesign.Manager` got a new Xcode Archive based constructor: `NewManagerWithArchive`


codesign/codesign.go
```diff
@@ -97,6 +124,11 @@ type Project interface {
 }
 
 func (m *Manager) getProject() (Project, error) {
+	// Manager is either initialised with an XCArchive or an Xcode Project.
+	if m.archive != nil {
+		return m.archive, nil
+	}
```
tothszabi, line 130: Do we need a separate variable for the archive? It is not used anywhere else than in this method. I think it is the same for `project` as well. If the archive conforms to the `Project` interface then we could assign it to the project variable.

codesign/archive.go
```diff
@@ -0,0 +1,52 @@
...
+
+// IsSigningManagedAutomatically ...
+func (a Archive) IsSigningManagedAutomatically() (bool, error) {
+	return a.archive.IsXcodeManaged(), nil
+}
+
+// Platform ...
+func (a Archive) Platform() (autocodesign.Platform, error) {
+	platformName := a.archive.Application.InfoPlist["DTPlatformName"]
+	switch platformName {
```
lpusok, line 30: You could extract the logic from ReadCodesignParameters, so it can be reused here.

codesign/archive.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return autocodesign.AppLayout{}, err
+	}
+	return *layout, nil
+}
+
+// ForceCodesignAssets ...
+func (a Archive) ForceCodesignAssets(distribution autocodesign.DistributionType, codesignAssetsByDistributionType map[autocodesign.DistributionType]autocodesign.AppCodesignAssets) error {
+	return nil
+}
```
tothszabi, line 52: I have a feeling that we do not have the best abstraction. Maybe it comes from the unfortunate `Project` naming which for me means `xcodeproj` and I would not even think about an archive. Also we cannot force set signing parameters as in a project so (in my mind) we only partially conform to the interface.  

For me something along the lines of `SigningDetailsProvidable` and `SigningDetailsSettable` (sorry for the Swift protocol names) would seem like a better separation than just to have a `Project` interface.

Let me just say that I do not have a better idea on really how I would do it. Everything probably requires a lot of refactoring and I am not sure yet my opinion fully settled. So in overall I am happy to leave it as well and just want to hear your opinion about it.

codesign/archive.go
godrei: Added `IosArchive.Platform` in https://github.com/bitrise-io/go-xcode/pull/141/commits/7eb170b841166e5182990b48779f8f51008a2db3

codesign/codesign.go
godrei: Merged project and archive in https://github.com/bitrise-io/go-xcode/pull/141/commits/7eb170b841166e5182990b48779f8f51008a2db3

codesign/archive.go
godrei: Separated `codesign.DetailsProvider` and `codesign.AssetWriter` in https://github.com/bitrise-io/go-xcode/pull/141/commits/7eb170b841166e5182990b48779f8f51008a2db3

lpusok (approved):

tothszabi (approved):

PR #146, opened 2021-12-08
Handle local and remote codesign asssets for one distribution type

ofalvai:
### Context

Merging local and remote assets does not work properly when there are multiple export methods requested (such as development and ad-hoc), because the logic does not differentiate assets based on the export method.

### Changes
- Ensure assets for each export method in a loop
- Thanks to this change, a lot of logic that worked on a `map[DistributionType]AppCodesignAssets` is simplified to a single `AppCodesignAssets` parameter

### DIFF
--- a/autocodesign/autocodesign.go
+++ b/autocodesign/autocodesign.go
@@ -187,7 +187,7 @@ func (m codesignAssetManager) EnsureCodesignAssets(appLayout AppLayout, opts Cod
 	codesignAssetsByDistributionType := map[DistributionType]AppCodesignAssets{}
 
 	for _, distrType := range distrTypes {
-		localCodesignAssets, missingCodesignAssets, err := m.localCodeSignAssetManager.FindCodesignAssets(appLayout, distrType, certsByType, devPortalDeviceUDIDs, opts.MinProfileValidityDays)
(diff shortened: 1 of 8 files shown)


autocodesign/autocodesign.go
```diff
@@ -184,41 +184,49 @@ func (m codesignAssetManager) EnsureCodesignAssets(appLayout AppLayout, opts Cod
...
 
-	localCodesignAssets, missingCodesignAssets, err := m.localCodeSignAssetManager.FindCodesignAssets(appLayout, distrTypes, certsByType, devPortalDeviceUDIDs, opts.MinProfileValidityDays)
-	if err != nil {
-		return nil, fmt.Errorf("failed to collect local code signing assets: %w", err)
-	}
+	codesignAssetsByDistributionType := map[DistributionType]AppCodesignAssets{}
 
-	printExistingCodesignAssets(localCodesignAssets)
+	for _, distrType := range distrTypes {
+		localCodesignAssets, missingCodesignAssets, err := m.localCodeSignAssetManager.FindCodesignAssets(appLayout, distrType, certsByType, devPortalDeviceUDIDs, opts.MinProfileValidityDays)
```
godrei, line 190: missingCodesignAssets is an AppLayout, IMO missingAppLayout would better describe this var

godrei (approved):

PR #147, opened 2021-12-17
Better ISO 8601 time parsing

tothszabi:
We had an issue reported about failing to parse provisioning profiles: 
https://github.com/bitrise-steplib/steps-ios-auto-provision-appstoreconnect/issues/92

A quick fix was deployed to mitigate the error and this is a follow up improvement.

At first look the cause was that the returned time format has changed from 
```
2006-01-02T15:04:05.000-0700
```
to
```
2006-01-02T15:04:05.000-07:00
```
(there is an extra `:` delimiter).

The root cause is actually different. The returned dates are in the ISO 8601 format. In this standard some formats are interchangeable and equivalent to each other. The best example is this:
- 2006-01-02T15:04:05.000Z
- 2006-01-02T15:04:05.000+00:00
- 2006-01-02T15:04:05.000+0000
- 2006-01-02T15:04:05.000+00

represent the same date. The time parser needs to handle all of them because (based on the standard description) they can be interchanged freely. It can happen that a time library / framework written in one language sends one format and one written in another language uses a different format. 

Go has inbuilt support for this via `time.RFC3339` but that only handles the zero time offset (Z) and the `[+/-]07:00` formats which means we need to manually try to parse the dates with the other formats as well.

### DIFF
--- a/autocodesign/devportalclient/appstoreconnect/time.go
+++ b/autocodesign/devportalclient/appstoreconnect/time.go
@@ -1,6 +1,7 @@
 package appstoreconnect
 
 import (
+	"fmt"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


Bence1001 (approved):
Excellent find and explanation, thanks Szabi! 👍

lpusok (approved):

PR #148, opened 2022-01-04
Fix for device registration regression

lpusok:
### Checklist
- [x] I've read and followed the [Contribution Guidelines](https://github.com/bitrise-steplib/.github/blob/main/CONTRIBUTING.md)
- [x] `step.yml` and `README.md` is updated with the changes (if needed)

### Version
Requires a *PATCH* [version update](https://semver.org/)

### Context

errors.As was incorrectly used:
```
	rerr := appstoreconnect.ErrorResponse{} <-- & missing
		if ok := errors.As(err, &rerr); ok {
```

Resolves: https://bitrise.atlassian.net/browse/STEP-1731

### Changes

- Correctly detect  appstoreconnect.ErrorResponse.
- Fixed a bug when an incorrect device is encountered, other devices still sill be registered instead of returning early.
- Added retry for fetching a Bundle ID from a profile, if the profile has been removed.

### DIFF
--- a/autocodesign/devportalclient/appstoreconnectclient/devices.go
+++ b/autocodesign/devportalclient/appstoreconnectclient/devices.go
@@ -64,9 +64,9 @@ func (d *DeviceClient) RegisterDevice(testDevice devportalservice.TestDevice) (*
 
 	registeredDevice, err := d.client.Provisioning.RegisterNewDevice(req)
 	if err != nil {
-		rerr := &appstoreconnect.ErrorResponse{}
-		if ok := errors.As(err, &rerr); ok {
-			if rerr.Response != nil && rerr.Response.StatusCode == http.StatusConflict {
+		var respErr *appstoreconnect.ErrorResponse
+		if ok := errors.As(err, &respErr); ok {
+			if respErr.Response != nil && respErr.Response.StatusCode == http.StatusConflict {
 				return nil, appstoreconnect.DeviceRegistrationError{
 					Reason: fmt.Sprintf("%v", err),
 				}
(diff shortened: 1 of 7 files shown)


autocodesign/devportalclient/appstoreconnectclient/devices.go
```diff
@@ -64,7 +64,7 @@ func (d *DeviceClient) RegisterDevice(testDevice devportalservice.TestDevice) (*
 
 	registeredDevice, err := d.client.Provisioning.RegisterNewDevice(req)
 	if err != nil {
-		rerr := appstoreconnect.ErrorResponse{}
+		rerr := &appstoreconnect.ErrorResponse{}
```
tothszabi, line 67: In the `profiles.go` file you were using the 
```
var respErr *appstoreconnect.ErrorResponse
```
format at both places and I wonder if we could use the same here as well just for consistency.

autocodesign/devportalclient/appstoreconnectclient/profiles_test.go
```diff
@@ -191,3 +190,36 @@ func newResponse(t *testing.T, status int, body map[string]interface{}) *http.Re
...
+					StatusCode: http.StatusNotFound,
+				},
+			},
+			wantErrType: autocodesign.ProfilesInconsistentError{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := wrapInProfileError(tt.err)
+			if tt.wantErrType == nil {
```
tothszabi, line 218: Do we need this if logic here if none of the test cases a nil `wantErrType` value? I would remove it and later add it back if it will be ever needed.

tothszabi (approved):

godrei (approved):

PR #150, opened 2022-01-19
Intalling certificate for local codesign assets

lpusok:
### Checklist

- [x] I've read and followed the [Contribution Guidelines](https://github.com/bitrise-steplib/.github/blob/main/CONTRIBUTING.md)
- [x] `step.yml` and `README.md` is updated with the changes (if needed)

### Version

Requires a *PATCH* [version update](https://semver.org/)

### Context

Make sure the certificate is installed when a matching local profile is found.

Resolves: https://bitrise.atlassian.net/browse/STEP-1715

### Changes

* Install certificate for locally found provisioning profiles. An alternative could have been to check weather certificate is already installed in the Keychain, always installing is the easier solution.
There was no check earlier that the selected certificate (or any certificate in the profile) was installed. This could have led to an error:
  - A Step with automatic code signing installs **Profile1**, this induces **Cert1** and **Cert2**. **Cert1** is selected (to be set as signing certificate in the project). **Profile1** is added, **Cert1** is added to the Keychain. (**Cert2** is not added.)
  - `xcodebuild archive` runs successfully
  - Another Step with automatic code signing is run. This time the previously added **Profile1** is found. However now **Cert2** is selected (randomly), it is not installed due to missing check.
   - `xcodebuild archive` fails with error: 
   
 ```
❌  error: No certificate for team 'AAABBBCCC' matching 'dcdcdc' found: Select a different signing certificate for CODE_SIGN_IDENTITY, a team that matches your selected certificate, or switch to automatic provisioning. (...)
```

### DIFF
--- a/autocodesign/autocodesign.go
+++ b/autocodesign/autocodesign.go
@@ -179,7 +179,6 @@ func (m codesignAssetManager) EnsureCodesignAssets(appLayout AppLayout, opts Cod
 			devPortalDeviceIDs = append(devPortalDeviceIDs, devPortalDevice.ID)
 			devPortalDeviceUDIDs = append(devPortalDeviceUDIDs, devPortalDevice.Attributes.UDID)
 		}
-
 	}
 
 	codesignAssetsByDistributionType := map[DistributionType]AppCodesignAssets{}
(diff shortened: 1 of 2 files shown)


shams-ahmed (dismissed):

ofalvai (approved):
LGTM, kudos for the extra tests 🚀

PR #152, opened 2022-01-24
Fix Xcode version parsing when output has warnings

ofalvai:
### Context

Xcode 13.2.1 has a known issue: https://developer.apple.com/forums/thread/698628

The additional lines in output breaks the version parsing.

### Changes

- Make the version parsing more robust
- Add more tests

### DIFF
--- a/utility/utility.go
+++ b/utility/utility.go
@@ -15,8 +15,20 @@ func getXcodeVersionFromXcodebuildOutput(outStr string) (models.XcodebuildVersio
 		return models.XcodebuildVersionModel{}, fmt.Errorf("failed to parse xcodebuild version output (%s)", outStr)
 	}
 
-	xcodebuildVersion := split[0]
-	buildVersion := split[1]
+	firstLineIndex := -1
+	for i, line := range split {
+		if strings.HasPrefix(line, "Xcode ") {
+			firstLineIndex = i
+			break
+		}
+	}
+
+	if firstLineIndex < 0 {
+		return models.XcodebuildVersionModel{}, fmt.Errorf("couldn't find Xcode version in output: %s", outStr)
+	}
+
+	xcodebuildVersion := split[firstLineIndex]
+	buildVersion := split[firstLineIndex+1]
 
 	split = strings.Split(xcodebuildVersion, " ")
 	if len(split) != 2 {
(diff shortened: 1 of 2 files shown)


utility/utility.go
```diff
@@ -15,8 +15,20 @@ func getXcodeVersionFromXcodebuildOutput(outStr string) (models.XcodebuildVersio
...
+	for i, line := range split {
+		if strings.HasPrefix(line, "Xcode ") {
+			firstLineIndex = i
+			break
+		}
+	}
+
+	if firstLineIndex < 0 {
+		return models.XcodebuildVersionModel{}, fmt.Errorf("couldn't find Xcode version in output: %s", outStr)
+	}
```
tothszabi, line 28: The context is not immediately visible when I am reading these lines. Looking at the test definitely help explaining why this is needed. What if we move this logic into a separate method with a more explanatory name like
```
func filterXcodeWarningsFromTheOutputAsAWorkaround(input []string) ([]string, error) {
    ...
}
```
and then at the call site we would only have 
```
split, err := filterXcodeWarningsFromTheOutputAsAWorkaround(split)
if err != nil {
    return models.XcodebuildVersionModel{}, err
}
```
This way the original method maybe remains more readable.

xcarchive/ios_test.go
```diff
@@ -24,7 +24,10 @@ func sampleRepoPath(t *testing.T) string {
 		require.NoError(t, err)
 		sampleArtifactsGitURI := "https://github.com/bitrise-samples/sample-artifacts.git"
 		cmd := command.New("git", "clone", sampleArtifactsGitURI, dir)
-		require.NoError(t, cmd.Run())
+		_, err = cmd.RunAndReturnTrimmedCombinedOutput()
+		if err != nil {
+			t.Errorf("git clone failed: %s", err)
+		}
```
tothszabi, line 30: It is not clear for me why this changes is needed. Could you tell me why?

ofalvai: Tests are failing in CI with the CI git account. The old code didn't provide a meaningful error message, I had to add this to see the root cause. But it's still work in progress, please ignore for now :)

utility/utility.go
ofalvai: Thanks, I also like this style better. Please take a look now.

tothszabi: Looks great.

tothszabi (approved):

PR #135, opened 2021-11-23
[MOB-231] local code signing assets

tothszabi:
### Context

The current `autocodesign` package pulls and creates the code signing assets needed for the given project, without taking local code signing assets into consideration.

This PR introduces `localcodesignasset` package, which collects the locally existing code signing assets for the given projects, and reports the missing assets.

`autocodesign.CodesignAssetManager` now utilises `localcodesignasset.Manager` to find existing assets and pulls and creates only the missing ones.

### Changes

- introduce `localcodesignasset.Manager` and wire in to `CodesignAssetManager.EnsureCodesignAssets`
- modify `autocodesign.EnsureTestDevices` to return `[]appstoreconnect.Device`, so that on caller side we can decide if device ids or device udids are needed
- the logic which selects a certificate for code signing is moved to `SelectCertificate` function, so that it can be reused
- `createWildcardBundleID` became exported to be able to use in `localcodesignasset` package
- improved profile type parsing for `profileutil.NewProvisioningProfileInfo`

### Investigation details

KNOWN ISSUE: `localcodesignasset` package matches the given target's entitlements file to Provisioning Profile entitlements, to check if all the target capabilities are included in the profile. 
If the target entitlements contains env vars, the package can not find the matching profile (with entitlements with resolved env vars).

PR #159, opened 2022-02-28
Fix parsing of macOS profile types

ofalvai:
### Context

There is a latent bug in parsing provisioning profiles that only affects macOS profiles. The profile type string value should have always been `osx` instead of `macos`, all the profiles I tested contain `osx` as the platform value.

This bug had been hidden by the parsing logic until we changed that: https://github.com/bitrise-io/go-xcode/pull/135/files#diff-bce396a3a6fcf20dea5ebb455b48500d48eb91b11baef46a7506c217d72b82b2

Note that the old code assigned macOS as the platform type by default and only changed it to iOS if it matched that platform. The new code does proper pattern matching for every possible platform, so it with `unknown platform type: macos` when parsing a macOS profile.

This is the root cause of https://github.com/bitrise-steplib/steps-export-xcarchive-mac/issues/13, because we recently bumped `go-xcode` in the step despite a failing CI check: https://github.com/bitrise-steplib/steps-export-xcarchive-mac/pull/12

It also affects the `xcode-archive-mac` step as `go-xcode` was bumped there too: https://github.com/bitrise-steplib/steps-xcode-archive-mac/pull/42

### DIFF
--- a/profileutil/info_model_test.go
+++ b/profileutil/info_model_test.go
@@ -3,6 +3,8 @@ package profileutil
 import (
 	"testing"
 
+	"github.com/fullsailor/pkcs7"
+
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


tothszabi (approved):

lpusok:
NOTE: could add a test using a sample macos archive from https://github.com/bitrise-io/sample-artifacts.

lpusok (approved):

PR #161, opened 2022-03-18
Improve empty passphrase handling

ofalvai:
### Context

Trying to clean up and unify the handling of certificates without a passphrase (which we allow uploading at the moment, so we should at least handle these certs consistently across all steps).

Related bug: https://github.com/bitrise-steplib/bitrise-step-manage-ios-code-signing/issues/8

### Changes

First of all, there is no real bug fix in this PR. It turned out that this code in go-xcode already supports parsing of the list inputs with empty passphrase items. The problem is on the step side where we marked the passphrase list input as required, so a single item of empty passphrase is an invalid step input.

This PR improves the following:

- Add tests for input parsing to make it explicit that we support empty passphrases
- Improve log messages here and there

### DIFF
--- a/codesign/inputparse_test.go
+++ b/codesign/inputparse_test.go
@@ -1,0 +1,129 @@
+package codesign
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/bitrise-io/go-xcode/v2/autocodesign/certdownloader"
+)
+
+func TestParseCertificates(t *testing.T) {
+	tests := []struct {
+		name    string
+		input   Input
+		want    []certdownloader.CertificateAndPassphrase
+		wantErr bool
+	}{
+		{
+			name: "One certificate and passphrase",
+			input: Input{
+				CertificateURLList:        "https://example.com/storage/development.p12 ",
+				CertificatePassphraseList: "password",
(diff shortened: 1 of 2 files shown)


shams-ahmed (approved):

godrei (approved):

codesign/inputparse_test.go
```diff
@@ -0,0 +1,129 @@
...
+			input: Input{
+				CertificateURLList:        "https://example.com/storage/development.p12|",
+				CertificatePassphraseList: "password|",
+				KeychainPath:              t.TempDir(),
+				KeychainPassword:          "keychainpassword",
+			},
+			wantErr: true, // Because empty passphrase is a valid value for a no-passphrase-cert, so the list has 2 items
+		},
+		{
+			name: "No certificate nor passphrase",
```
lpusok, line 60: We may want to allow this in the future (Xcode Cloud Signing can use distribution certs without them being specified here)

ofalvai: Fair point, I didn't think about that. But if we decide to support Cloud Signing in every step, it's going to require many other changes as well, for example, every step should allow an empty certificate URL too, not just an empty passphrase.

PR #167, opened 2022-05-02
Avoid eagerly parsing every target in project

ofalvai:


xcodeproject/xcodeproj/xcodeproj_test.go
```diff
@@ -389,15 +389,10 @@ func TestTargets(t *testing.T) {
 		target, ok := project.Proj.Target("7D0342D720F4B5AD0050B6A6")
 		require.True(t, ok)
 
-		dependentTargets := target.DependentTargets()
+		dependentTargets := project.DependentTargetsOfTarget(target)
 		require.Equal(t, 2, len(dependentTargets))
 		require.Equal(t, "WatchKitApp", dependentTargets[0].Name)
 		require.Equal(t, "WatchKitApp Extension", dependentTargets[1].Name)
-
```
ofalvai, line 396: Deleted this asserts because `DependentExecutableProductTargets()` is removed. Also, these asserts were testing the same as the section above.

xcodeproject/xcodeproj/target_dependency_test.go
```diff
@@ -1,78 +0,0 @@
-package xcodeproj
```
ofalvai, line 1: This test is no longer relevant as we no longer parse the target dependencies, only an ID reference. That is tested in `target_test.go`

xcodeproject/xcodeproj/target.go
```diff
@@ -31,49 +31,16 @@ type Target struct {
...
-		if childTarget.ID == targetID {
+		if targetDependency.TargetID == targetID {
 			return true
 		}
 	}
 	return false
 }
 
-// DependentExecutableProductTargets ...
-func (t Target) DependentExecutableProductTargets() []Target {
```
ofalvai, line 60: This was only used at 2 places, so I decided to inline at the callsites. This method was basically the same as `DependentTargets()` with an extra filtering.

xcodeproject/xcodeproj/target.go
```diff
@@ -31,49 +31,16 @@ type Target struct {
 	buildPhaseIDs          []string
 }
 
-// DependentTargets ...
-func (t Target) DependentTargets() []Target {
```
ofalvai, line 35: Moved to `xcodeproj.go`

xcodeproject/xcodeproj/target.go
```diff
@@ -107,22 +74,11 @@ func (t Target) IsUITestProduct() bool {
...
 }
 
-func (t Target) isAppClipProduct() bool {
+// IsAppClipProduct ...
+func (t Target) IsAppClipProduct() bool {
 	return t.ProductType == appClipProductType
 }
 
-// CanExportAppClip ...
-func (t Target) CanExportAppClip() bool {
```
ofalvai, line 115: Its only usage is in `bitrise-init`, so I'm rather deleting this and making `IsAppClipProduct()` public. The code in `bitrise-init` already has access to the project, so it can call `Xcodeproj.DependentTargetsOfTarget()` and iterate over the results.

xcodeproject/xcodeproj/target_dependency.go
```diff
@@ -19,13 +19,8 @@ func parseTargetDependency(id string, objects serialized.Object) (TargetDependen
 		return TargetDependency{}, err
 	}
 
-	target, err := parseTarget(targetID, objects)
```
lpusok, line 22: Can you add a test case with more dependent targets?
Just want to be sure that it is guaranteed that:
- We parse all dependent targets, even with this part removed.
- If there is a target dependency that is missing from the project, it is handled gracefully. (At least no additional errors are returned compared to now)

xcodeproject/xcodeproj/target.go
ofalvai: Good point! But isn't the old implementation (potentially) incorrect because it skips certain targets, while the new implementation returns with all the possible executable targets?

I'm fine with re-adding this method to keep behavior the same, but I don't see any harm in this behavior change as it returns more, not fewer targets.

xcodeproject/xcodeproj/target.go
ofalvai: Fair point! If the main concern is testability, is it enough if I extract this logic into a testable private method in `bitrise-init`, or you'd rather leave this method in `go-xcode`?

xcodeproject/xcodeproj/target_dependency.go
```diff
@@ -2,10 +2,10 @@ package xcodeproj
 
 import "github.com/bitrise-io/go-xcode/xcodeproject/serialized"
 
-// TargetDependency ...
+// TargetDependency is a reference to another Target that is a dependency of a given Target
 type TargetDependency struct {
-	ID     string
-	Target Target
+	ID       string
+	TargetID string
```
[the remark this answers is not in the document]
ofalvai: It's not used anywhere else, so I made it private now. I want to keep the structure the same, to mirror the data structures of `pbxproj`.

xcodeproject/xcodeproj/target_dependency.go
ofalvai: For the first point, take a look at `target_test.go`, it has tests for parsing the entire Target structure with the dependent target ID reference as well.

Regarding the rest of the parsing and any missing dependency: It's handled now in `Xcodeproj.DependentTargetsOfTarget()`, which is tested in `xcodeproj_test.go` with 2 dependencies. I also added a new test case for a target that has no dependencies. I'm not sure about the case of an invalid dependency, I think this setup would fail anyway sooner or later in `xcodebuild`. Our old implementation produced an error (target.go line 127), now it returns an empty dependency list. To me it seems a better solution, but I can change my mind.

shams-ahmed:
Had a look at the changes, seem good so far. 
I can't see any issues apart from adding additional tests.

xcodeproject/xcodeproj/xcodeproj.go
```diff
@@ -106,6 +106,24 @@ func (p XcodeProj) TargetInfoplistPath(target, configuration string) (string, er
...
 }
 
+// DependentTargetsOfTarget returns with all dependencies of a given target, including the transitive dependencies.
+// The returned list contains each target only once, using the target ID for uniqueness.
+func (p XcodeProj) DependentTargetsOfTarget(target Target) []Target {
+	var dependentTargets []Target
+	for _, dependency := range target.Dependencies {
+		childTarget, ok := p.Proj.Target(dependency.TargetID)
+		if !ok {
+			log.Warnf("couldn't find dependency %s of target %s (%s), skipping", dependency.TargetID, target.Name, target.ID)
```
lpusok, line 116: You probably want to add a continue line after the warning.

ofalvai: Whoops, fixed!

xcodeproject/xcodeproj/target.go
lpusok: After some discussion I now think it is a good idea to deduplicate the recursive iteration; as we want to make it more robust in the next PR (to reduce extraneous iterations).
There is no obvious use case where new executable targets would be returned even with the changed logic, so we can go ahead with this.

xcodeproject/xcodeproj/target_dependency.go
lpusok: I think it does return a partial dependency list in case of an invalid dependency, right?
Not returning an error is a good idea, as Xcode projects can have missing targets sometimes (Xcode has some logic that recreates some targets if not found).

xcodeproject/xcodeproj/target.go
lpusok: A consideration was to make it harder to misuse this API (like using IsAppClipProduct instead of CanExportAppClip).
But as there is one use I am ok with removing it.

lpusok (approved):

gabo76555:
https://github.com/bitrise-io/xcode-project/tree/master

PR #169, opened 2022-05-04
[NOJIRA] Remove xcodeproj package

lpusok:
No repos (other than forked private ones) depend on this package: https://cs.github.com/?scopeName=All+repos&scope=&q=%22github.com%2Fbitrise-io%2Fgo-xcode%2Fxcodeproj%5C%22%22

Decided to remove as the tuist project generator needs some maintenance with new Xcode versions.
A still valid use case for tuist is to generate sample apps tested in the CI.


ofalvai (dismissed):
Sweet! LGTM, but I'd rather have someone else taking a look too!

shams-ahmed (dismissed):

ofalvai (approved):

PR #170, opened 2022-05-05
Allow passing `xcconfig` file path instead of file content

ofalvai:
### Context

Feature request for the Xcode Archive step: https://github.com/bitrise-steplib/steps-xcode-archive/issues/273

### Changes

Allow passing a file path to an existing `xcconfig` file, instead of providing the file contents directly.


xcconfig/xcconfig.go
```diff
@@ -2,34 +2,52 @@ package xcconfig
...
 
-func (w writer) Write(content string) (string, error) {
+func (w writer) Write(input string) (string, error) {
+	if w.isPath(input) {
+		pathExists, err := w.pathChecker.IsPathExists(input)
+		if err != nil {
+			return "", fmt.Errorf(err.Error())
+		}
+		if !pathExists {
+			return "", fmt.Errorf("provided xcconfig file path doesn't exist")
```
ofalvai, line 35:
```suggestion
			return "", fmt.Errorf("provided xcconfig file path doesn't exist: %s", input)
```

Let's include the actual path in the error message

xcconfig/xcconfig_test.go
```diff
@@ -27,3 +29,33 @@ func Test_WhenWritingXCConfigContent_ThenItShouldReturnFilePath(t *testing.T) {
 		assert.Equal(t, expectedPath, path)
 	}
 }
+
+func Test_XCConfigInput_NonExistentPathErrors(t *testing.T) {
```
ofalvai, line 33: nit: I would follow the `// Given/When/Then` grouping convention in the new test cases too for consistency.

xcconfig/xcconfig.go
```diff
@@ -2,34 +2,52 @@ package xcconfig
...
+
 	"github.com/bitrise-io/go-utils/v2/fileutil"
 	"github.com/bitrise-io/go-utils/v2/pathutil"
-	"path/filepath"
 )
 
 // Writer ...
 type Writer interface {
-	Write(content string) (string, error)
+	Write(input string) (string, error)
```
ofalvai, line 14: This would be the perfect time to add a few lines of documentation to this interface and method. At the moment it's not clear what the `input` string can contain (file content or path) and what happens in each case.

ofalvai (commented):
Looks really good @sujitpoudel 🚀 
I have only a few minor remarks.

xcconfig/xcconfig_test.go
sujitpoudel: Cool. Fixed it up.

sujitpoudel (approved):

PR #173, opened 2022-05-09
Added fallback option

lpusok:
### Context

Resolves: https://bitrise.atlassian.net/browse/STEP-1932

### Changes

- Added profiledownloader
- Added manual profiles fallback on code signing error


autocodesign/localcodesignasset/profile.go
```diff
@@ -14,6 +16,23 @@ type Profile struct {
 	certificateIDs []string
 }
 
+// NewProfile wraps a local profile in the autocodesign.Profile interface
+func NewProfile(info profileutil.ProvisioningProfileInfoModel, content []byte) autocodesign.Profile {
```
lpusok, line 20: Moved from profielconverter, so it can be reused on its own.

codesign/inputparse.go
```diff
@@ -92,3 +105,52 @@ func validateCertificates(certURLList string, certPassphraseList string) ([]stri
 func splitAndClean(list string, sep string, omitEmpty bool) (items []string) {
 	return sliceutil.CleanWhitespace(strings.Split(list, sep), omitEmpty)
 }
+
+func parseFallbackProvisioningProfiles(profilesList string) ([]string, error) {
```
godrei, line 109: so this returns a clean, valid list of profile file paths, based on a profile list (that contains file profile file paths and dirs storing profiles).

I propose updating the function name and adding some description, something like:

```
// validateAndExpandProfilePaths validates and expands profilesList.
// For directory list items, the contained profiles' path will be returned.
validateAndExpandProfilePaths
```

codesign/inputparse.go
```diff
@@ -92,3 +105,52 @@ func validateCertificates(certURLList string, certPassphraseList string) ([]stri
...
+		for _, dirEntry := range dirEntries {
+			if dirEntry.Type().IsDir() || !dirEntry.Type().IsRegular() {
+				continue
+			}
+
+			if strings.HasSuffix(dirEntry.Name(), codesignasset.ProfileIOSExtension) {
+				profileURL := fmt.Sprintf("file://%s", filepath.Join(profilesDir, dirEntry.Name()))
+				validProfiles = append(validProfiles, profileURL)
+			}
+		}
```
godrei, line 152: this could be moved to a separate function, to make this function cleaner.

autocodesign/autocodesign.go
```diff
@@ -110,6 +112,18 @@ type CertificateProvider interface {
 // LocalCertificates is a map from the certificate type (development, distribution) to an array of installed certs
 type LocalCertificates map[appstoreconnect.CertificateType][]certificateutil.CertificateInfoModel
 
+// LocalProfile ...
+type LocalProfile struct {
+	APIProfile Profile
+	Info       profileutil.ProvisioningProfileInfoModel
+}
+
+// ProfileProvider returns manually generated profiles, used as a fallback
```
godrei, line 121: why manually generated profiles only? I think how other code uses this interface should not be here.

codesign/inputparse.go
lpusok: Updated.

codesign/inputparse.go
lpusok: Agree, updated.

autocodesign/autocodesign.go
lpusok: Made less specific.

godrei (approved):

PR #176, opened 2022-05-23
fix build issue with XCPretty failing to pass logger object

shams-ahmed:
- fix build issue with XCPretty failing to pass logger object
- update all dependencies 
- fix tests from outdated mock class


xcpretty/xcpretty.go
```diff
@@ -122,7 +122,9 @@ func (x xcpretty) IsInstalled() (bool, error) {
 		return false, err
 	}
 
-	return ruby.NewEnvironment(factory, locator).IsGemInstalled("xcpretty", "")
+	logger := log.NewLogger()
```
lpusok, line 125: To follow the patterns used, it would make more sense to add logger as a param in NewXcpretty() (even if it is a breaking change).

shams-ahmed: updated!

lpusok (approved):

PR #179, opened 2022-06-10
Add Test Plan support to Scheme

godrei:
### Context

This PR adds Test Plan support to `xcodebuild` and `xcodeproj` packages.

### Changes

**`xcodebuild` package:**
- Add test plan property and setter to `CommandBuilder`

---

**`xcodeproj` package:**
- add `TestPlans` property to the Scheme's `TestAction`
- add utility functions on `TestPlanReference`
  - `TestPlanReference.IsDefault` and `TestPlanReference.Name`
- add `DefaultTestPlan` utility func to `Scheme`

---

**Non-function changes (`xcodebuild` package):**
- [BREAKING CHANGE] `NewCommandBuilder` constructor parameters modified:
  - `isWorkspace` parameter removed, as it can be simply calculated based on the `projectPath`
  - `action` parameter made a variadic parameter, so that multiple action can be specified without using `SetCustomBuildAction`
  - `SetCustomBuildAction` removed as being redundant
  - order of args modified in the generated command: action -> options -> build settings
  - custom option handling based on the action is removed, as it was not thorough anyway
- `CommandBuilder` test modified to use a function to build the CommandBuilder, this way the tests are closer to the client side usage
- Scheme test changes:
  - Merge `TestOpenScheme`, `TestAppBuildActionEntry` and `TestAppTestActionEntry` in the new `Test_GivenSimpleScheme_WhenOpens`, so that `ios-simple-objc.xcscheme` gets opened only once
  - `TestScheme_Marshal` renamed to `TestScheme_Marshal`


hisaac (approved):
Code changes look good to me!

I have a couple questions, but will approve anyway.

- I see you're merging this into the `v1` branch. Is that intended?
- Since we're moving the xcscheme XML test data to its own file, would it make sense to include actual .xcscheme files instead of having these as Go strings? Maybe would add a little more overhead for testing, but it might be a little more accurate? I'm fine either way on this, just wondering what you think.

godrei:
> I see you're merging this into the v1 branch. Is that intended?

It is intentional, on master (v2), we do not have the `xcodeproj` package. 
`xcodebuild` package exists, but functionality is not migrated.

xcodeproject/xcscheme/xcscheme_test.go
```diff
@@ -52,157 +52,22 @@ func TestAppBuildActionEntry(t *testing.T) {
...
 }
 
-func TestAppTestActionEntry(t *testing.T) {
-	var scheme Scheme
-	require.NoError(t, xml.Unmarshal([]byte(schemeContent), &scheme))
-
+func assertIosSimpleObjCSchemeTestActionEntry(scheme Scheme, t *testing.T) {
 	require.Equal(t, "Debug", scheme.TestAction.BuildConfiguration)
-	require.Equal(t, 2, len(scheme.TestAction.Testables))
+	require.Equal(t, 1, len(scheme.TestAction.Testables))
```
godrei, line 64: On the [master branch](https://github.com/bitrise-io/sample-apps-ios-simple-objc/blob/master/ios-simple-objc/ios-simple-objc.xcodeproj/xcshareddata/xcschemes/ios-simple-objc.xcscheme#L44-L55) the scheme has 1 testable.

godrei:
> Since we're moving the xcscheme XML test data to its own file, would it make sense to include actual .xcscheme files instead of having these as Go strings? Maybe would add a little more overhead for testing, but it might be a little more accurate? I'm fine either way on this, just wondering what you think.

Updated the code to use real scheme files.
Since `scheme.Open` works with actual files, using real scheme files, made the tests more clear I think.
I did some updates to open the test files less times, also updated the PR description with the test related changes.

hisaac (approved):
LGTM! 🚀

PR #183, opened 2022-06-20
Moved simulator package to v2

lpusok:
Moved from:
- simulator package from [Xcode test Step](https://github.com/bitrise-steplib/steps-xcode-test/blob/d084e66c1726162ca4f06ae6ccbabccf8db2e701/simulator/simulator.go#L1)
- BootSimulator from _[v1/simulator](https://github.com/bitrise-io/go-xcode/blob/600183635828325b5297c47a41c365d6a5947332/simulator/simulator.go#L284)_ 
- Added _Erase_ and _WaitForBootFinished_


simulator/simulator.go
```diff
@@ -0,0 +1,260 @@
...
+	cmd := m.commandFactory.Create("xcrun", []string{"simctl", "shutdown", id}, &command.Opts{
+		Stdout: os.Stdout,
+		Stderr: os.Stderr,
+	})
+
+	m.logger.Donef("$ %s", cmd.PrintableCommandArgs())
+	exitCode, err := cmd.RunAndReturnExitCode()
+	if err != nil {
+		if errorutil.IsExitStatusError(err) {
+			if exitCode == 149 { // Simulator already shut down
```
matrangam, line 222: This comment is probably enough for this particular case, do you think it's reasonable to add some constants for these exit codes?

simulator/simulator.go
```diff
@@ -0,0 +1,260 @@
...
+	"strings"
+	"time"
+
+	"github.com/bitrise-io/go-utils/errorutil"
+	"github.com/bitrise-io/go-utils/v2/command"
+	"github.com/bitrise-io/go-utils/v2/log"
+)
+
+// Manager provides methods for issuing Simulator commands
+type Manager interface {
```
matrangam, line 18: Is a better name for this `iOSSimulatorManager`

lpusok: disagree, it will be used as _simulator.Manager_. no need to add the package name in the name.

matrangam: Ah good point.

simulator/simulator.go
lpusok: Extracted to constant.

matrangam (approved):

PR #185, opened 2022-09-11
[BE-282] Install missing device at runtime.

lpusok:
On Xcode 14 RC, iPhone 8 is not installed by default. This is the default device used in Steps.
xcode-test, start-xcode-simulator Steps will fail if a given device (`iPhone 8` with runtime`16.0`) is not available. As long as the runtime (`iOS 16.0`) is installed, we can add new devices at Step runtime, avoiding failures.

`Process config: simulator UDID lookup failed: device type (iPhone 8) with runtime OS (16.0) is unavailable` (https://app.bitrise.io/build/3f440c13-328b-451c-9692-5a4caebb1684)

## Changes
- Look up device type identifier, this is needed for the `xcrun simctl xreate` command.
- Check that a given runtime supports device.
- Use `xcrun simctl create` to add new device.


tothszabi (approved):
This is a great little addition. 

I checked that iOS 16 will be the 6th OS which still supports iPhone 8. It feels like it will be dropped soonish. We will not have this problem for the next year but it might worth bumping the default value of our steps to a more recent device version.

PR #188, opened 2022-09-16
Use standard linters

ofalvai:
Use the standard linters and config we use in other step repos.

This PR drops errcheck and golint in favor of golangci-lint through the check internal step.

There were a few lint violations with the new rules, this PR fixes those.

```
xcarchive/ios_test.go:68:9: composites: github.com/bitrise-io/go-xcode/xcarchive.IosExtension struct literal uses unkeyed fields (govet)
								{
								^
xcarchive/ios_test.go:93:8: composites: github.com/bitrise-io/go-xcode/xcarchive.IosExtension struct literal uses unkeyed fields (govet)
							{
							^
xcarchive/ios_test.go:104:8: composites: github.com/bitrise-io/go-xcode/xcarchive.IosExtension struct literal uses unkeyed fields (govet)
							{
							^
autocodesign/devportalclient/appstoreconnectclient/devices_test.go:18:9: copylocks: Do passes lock by value: github.com/bitrise-io/go-xcode/v2/autocodesign/devportalclient/appstoreconnectclient.MockDeviceClient contains github.com/stretchr/testify/mock.Mock contains sync.Mutex (govet)
func (c MockDeviceClient) Do(req *http.Request) (*http.Response, error) {
        ^
autocodesign/devportalclient/appstoreconnectclient/devices_test.go:42:38: copylocks: call of appstoreconnect.NewClient copies lock value: github.com/bitrise-io/go-xcode/v2/autocodesign/devportalclient/appstoreconnectclient.MockDeviceClient contains github.com/stretchr/testify/mock.Mock contains sync.Mutex (govet)
	client := appstoreconnect.NewClient(mockClient, "keyID", "issueID", []byte("privateKey"))
	                                    ^
autocodesign/autocodesign_test.go:278:28: errorsas: second argument to errors.As should not be *error (govet)
				(tt.wantErr != nil && !errors.As(err, &tt.wantErr)) {
				                       ^
autocodesign/projectmanager/projecthelper_test.go:106:16: SA5011: possible nil pointer dereference (staticcheck)
			if projHelp.Configuration != tt.wantConfiguration {
			            ^
autocodesign/projectmanager/projecthelper_test.go:103:7: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
			if projHelp == nil {
			   ^
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,26 @@
+run:
+  skip-dirs:
+  - .*/mocks
+
+issues:
+  # https://github.com/golangci/golangci-lint/issues/2439
+  exclude-use-default: false
+
+linters:
+  enable:
+  - errcheck
+  - gosimple
+  - govet
+  - ineffassign
+  - staticcheck
+  - typecheck
+  - unused
+  - revive
+
+linters-settings:
+  revive:
+    severity: error
+    rules:
+    - name: exported
+      arguments:
+      - checkPrivateReceivers
--- a/autocodesign/autocodesign_test.go
+++ b/autocodesign/autocodesign_test.go
@@ -184,7 +184,7 @@ func Test_codesignAssetManager_EnsureCodesignAssets(t *testing.T) {
 		appLayout AppLayout
 		opts      CodesignAssetsOpts
 		want      map[DistributionType]AppCodesignAssets
-		wantErr   error
+		wantErr   *DetailedError
 	}{
 		{
 			name: "no valid certs found",
--- a/autocodesign/certdownloader/certdownloader.go
+++ b/autocodesign/certdownloader/certdownloader.go
@@ -32,6 +32,7 @@ func NewDownloader(certs []CertificateAndPassphrase, client *http.Client) autoco
 	}
 }
 
+// GetCertificates ...
 func (d downloader) GetCertificates() ([]certificateutil.CertificateInfoModel, error) {
 	var certInfos []certificateutil.CertificateInfoModel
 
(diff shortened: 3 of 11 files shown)


.golangci.yml
```diff
@@ -0,0 +1,26 @@
...
+  - unused
+  - revive
+
+linters-settings:
+  revive:
+    severity: error
+    rules:
+    - name: exported
+      arguments:
+      - checkPrivateReceivers
```
tothszabi, line 26: This setting is exactly the same as in your other PR:
https://github.com/bitrise-io/go-utils/pull/168

Do we need to standardise it like what we are doing with the yamlint settings? The `steps-check` step contains the default settings and writes it to a file before running that given check.

ofalvai: Hard to say, my argument for duplicating it across the few library repos is that we only have 3 such repos at the moment and I don't expect to create new ones in the future. 

Are you thinking about creating a new step, such as `libs-check`, or extending the `steps-check` code to apply different golangci-lint rules for steps and libs?

tothszabi: I was thinking of somehow adding this config yaml to the `steps-check` step and somehow configuring when to apply it. But now that I think about it we can do it later as well and it will be easy to do it even if we merge this PR as it is.

One of the CI seems to have failed while the other succeeded so I think probably just restarting the build probably will solve it.

ofalvai: Agree, let's do the abstraction later. I restarted the CI build, it was indeed just flakiness. The only remaining thing is a ✅ from you 🥺 😁

tothszabi (approved):

PR #189, opened 2022-12-05
[CI-267] Add generic xcodebuild error finder

tothszabi:
It seems like the xcodebuild error formats are generic and shared between all of the xcodebuild subcommands. The xcode-test step was producing similar errors as the xcode-archive one. 

This PR extracts and adds test to the already existing custom xcodebuild error finder logic which at the moment only lives in the xcode-archive step: https://github.com/bitrise-steplib/steps-xcode-archive/blob/9f4ef188d017b110374ea71ba9445bae1acfa6ef/step/errors.go

### DIFF
--- a/errorfinder/errorfinder.go
+++ b/errorfinder/errorfinder.go
@@ -1,0 +1,80 @@
+package errorfinder
+
+import (
+	"bufio"
+	"strings"
+)
+
+// FindXcodebuildErrors ...
+func FindXcodebuildErrors(out string) []string {
+	var errorLines []string       // single line errors with "error: " prefix
+	var xcodebuildErrors []string // multiline errors starting with "xcodebuild: error: " prefix
+	var nserrors []nsError        // single line NSErrors with schema: Error Domain=<domain> Code=<code> "<reason>" UserInfo=<user_info>
+
+	isXcodebuildError := false
+	var xcodebuildError string
+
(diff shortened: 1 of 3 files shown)


godrei (approved):

errorfinder/errorfinder.go
```diff
@@ -0,0 +1,139 @@
+package errorfinder
+
+import (
+	"bufio"
+	"regexp"
+	"strings"
+)
+
+// NSError ...
+type NSError struct {
```
godrei, line 10: Do we need to expose this struct?

tothszabi: Good point. It is just for internal usage. I have removed it.

errorfinder/errorfinder.go
```diff
@@ -0,0 +1,80 @@
+package errorfinder
+
+import (
+	"bufio"
+	"strings"
+)
+
+// FindXcodebuildErrors ...
```
ofalvai, line 8: I'm not a fan of documenting every public method, but in this case, it would be great to quickly see an example input and the returned string list.

ofalvai (approved):

errorfinder/errorfinder.go
ofalvai: I'm okay with that, thanks for explaining the reasoning. (After writing my comment, I immediately noticed the test cases, those also serve as a kind of documentation :) )

PR #196, opened 2023-06-13
[CI-615] File based device parsing

tothszabi:
Bitrise does not provide a proper device management for the users. This PR does not aim to solve that but rather adding an extra easy way managing the test devices. 

The users will be able to provide a path to a file which contains a comma separated list of device identifiers. If this input is set then it will have priority over the Bitrise service connection. 

This parsing logic will be used by three steps:
- steps-xcode-archive
- steps-export-xcarchive
- steps-xcode-build-for-test

And here are some links to see how these changes would look in the steps:
- https://github.com/bitrise-steplib/steps-xcode-archive/compare/CI-615-file-based-device-list?expand=1
- https://github.com/bitrise-steplib/steps-export-xcarchive/compare/CI-615-file-based-device-list?expand=1
- https://github.com/bitrise-steplib/steps-xcode-build-for-test/compare/CI-615-file-based-device-list?expand=1

### DIFF
--- a/devportalservice/testdevice.go
+++ b/devportalservice/testdevice.go
@@ -1,0 +1,51 @@
+package devportalservice
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/bitrise-io/go-utils/pathutil"
+)
+
+// TestDevice ...
+type TestDevice struct {
+	ID     int `json:"id"`
+	UserID int `json:"user_id"`
+	// DeviceID is the Apple device UDID
+	DeviceID   string    `json:"device_identifier"`
+	Title      string    `json:"title"`
+	CreatedAt  time.Time `json:"created_at"`
+	UpdatedAt  time.Time `json:"updated_at"`
+	DeviceType string    `json:"device_type"`
+}
+
+// ParseTestDevicesFromFile ...
+func ParseTestDevicesFromFile(path string, currentTime time.Time) []TestDevice {
+	absPath, err := pathutil.AbsPath(path)
+	if err != nil {
+		return []TestDevice{}
+	}
+
+	bytes, err := os.ReadFile(absPath)
+	if err != nil {
+		return []TestDevice{}
+	}
+
+	fileContent := strings.TrimSpace(string(bytes))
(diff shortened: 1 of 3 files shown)


devportalservice/testdevice.go
```diff
@@ -0,0 +1,51 @@
...
+	// DeviceID is the Apple device UDID
+	DeviceID   string    `json:"device_identifier"`
+	Title      string    `json:"title"`
+	CreatedAt  time.Time `json:"created_at"`
+	UpdatedAt  time.Time `json:"updated_at"`
+	DeviceType string    `json:"device_type"`
+}
+
+// ParseTestDevicesFromFile ...
+func ParseTestDevicesFromFile(path string, currentTime time.Time) []TestDevice {
```
godrei, line 25: Shouldn't we propagate errors to the caller and let them decide how to handle it?

devportalservice/testdevice.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	fileContent := strings.TrimSpace(string(bytes))
+	identifiers := strings.Split(fileContent, ",")
+
+	var testDevices []TestDevice
+	for i, identifier := range identifiers {
+		testDevices = append(testDevices, TestDevice{
+			DeviceID:   identifier,
+			Title:      fmt.Sprintf("Device %d", i+1),
```
godrei, line 43: When you register devices on the Account Settings/Test Devices page you have the option to provide a name for the test device? Shouldn't we support this functionality?

devportalservice/testdevice.go
```diff
@@ -0,0 +1,51 @@
...
+			CreatedAt:  currentTime,
+			UpdatedAt:  currentTime,
+			DeviceType: "unknown",
```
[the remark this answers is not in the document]
tothszabi: I think this is only a Bitrise specific property and not used during the device registration. It is only used as part of a log message and that is all. So based on my debugging `unknown` will not affect anything.

devportalservice/testdevice.go
tothszabi: That is a good point. I will change that.

godrei (approved):

PR #197, opened 2023-07-05
Adding support for default Bitrise device.

lpusok:
### Context
We want to remove hardcoded device names (`iPhone 8 Plus`) from Step inputs.
On most Stacks there is a default device created, the Steps will use that by default. If the default device is not available, will use the first available device.

### DIFF
--- a/destination/device_finder.go
+++ b/destination/device_finder.go
@@ -8,6 +8,8 @@ import (
 	"github.com/bitrise-io/go-utils/v2/log"
 )
 
+const defaultDeviceName = "Bitrise iOS default"
+
 // Device is an available device
 type Device struct {
 	Name   string
--- a/destination/device_finder.go
+++ b/destination/device_finder.go
@@ -8,6 +8,7 @@ import (
 	"github.com/bitrise-io/go-utils/v2/log"
 )
 
+// Keep it in sync with https://github.com/bitrise-io/image-build-utils/blob/master/roles/simulators/defaults/main.yml#L14
 const defaultDeviceName = "Bitrise iOS default"
 
 // Device is an available device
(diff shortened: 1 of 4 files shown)


destination/device_finder.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/bitrise-io/go-utils/v2/log"
 )
 
+const defaultDeviceName = "Bitrise iOS default"
```
ofalvai, line 11:
```suggestion
// Keep it in sync with https://github.com/bitrise-io/image-build-utils/blob/master/roles/simulators/defaults/main.yml#L14
const defaultDeviceName = "Bitrise iOS default"
```

destination/parse.go
```diff
@@ -197,15 +197,28 @@ func (d deviceFinder) filterDeviceList(wantedDevice Simulator) (Device, error) {
...
 
-	deviceTypeIdentifier, err := d.lookupDeviceTypeID(wantedDevice.Name)
-	if err != nil {
-		return Device{}, err
+	// As the name of the device matches the device type ('iPhone 11') for factory created devices, look up device by name.
+	// If the default name is required and already created will use that.
+	for _, device := range devices {
+		if device.Name == wantedDevice.Name {
+			if !device.IsAvailable {
+				return Device{}, fmt.Errorf("device (%s) with runtime OS (%s) is unavailable: %s", wantedDevice.Name, runtime.Version, device.AvailabilityError)
```
ofalvai, line 205: Can't we continue with one of the fallback options instead of failing with an error?

lpusok: It is not worth it, this case only happens if a specific device is already created but this field contains an error (like: ``` "devices" : {
	    "com.apple.CoreSimulator.SimRuntime.watchOS-7-4" : [
	      {
	        "availabilityError" : "runtime profile not found",```)

ofalvai (approved):

PR #198, opened 2023-07-05
Supported platforms

godrei:
### Context

In the recent Xcode version when creating a new project a Multiplatform App is the default initial configuration. Such a project can be built for iOS and macOS platforms too.

These projects' `SDKROOT` target build setting is set to `auto` and has a `SUPPORTED_PLATFORMS` target build setting with `"iphoneos iphonesimulator macosx"` value. 

This setup is not considered an iOS project by the project scanner, because the `pathfilters.SDKFilter` expects `SDKROOT=iphoneos`.

This PR updates `pathfilters.SDKFilter` to use the `SUPPORTED_PLATFORMS` target build settings when deciding if the given project can be built with the given SDK or not.

JIRA ticket: https://bitrise.atlassian.net/browse/BIVS-1998

### Changes

- Update `pathfilters.SDKFilter` to utilize `SUPPORTED_PLATFORMS` target build settings
- Test `pathfilters.SDKFilter` on a Multiplatform App

### Investigation

- An example can be found here: https://github.com/bitrise-io/samples-ios-swiftui-bitrise-todos/blob/main/Bitrise%20TODOs%20Sample.xcodeproj/project.pbxproj#L481-L482
- Supported Platforms build setting reference: https://developer.apple.com/documentation/xcode/build-settings-reference#Supported-Platforms

### DIFF
--- a/pathfilters/pathfilters.go
+++ b/pathfilters/pathfilters.go
@@ -74,7 +74,7 @@ var AllowIphoneosSDKFilter = SDKFilter("iphoneos", true)
 var AllowMacosxSDKFilter = SDKFilter("macosx", true)
 
 // SDKFilter ...
-func SDKFilter(sdk string, allowed bool) pathutil.FilterFunc {
+func SDKFilter(expectedSDK string, allowed bool) pathutil.FilterFunc {
 	return func(pth string) (bool, error) {
 		found := false
 
@@ -127,6 +127,10 @@ func SDKFilter(sdk string, allowed bool) pathutil.FilterFunc {
 					sdkMap[sdk] = true
 				}
 
+				if sdk != "auto" {
+					continue
+				}
+
 				supportedPlatformsValue, err := buildConfiguration.BuildSettings.String("SUPPORTED_PLATFORMS")
 				if err == nil {
 					supportedPlatforms := strings.Split(supportedPlatformsValue, " ")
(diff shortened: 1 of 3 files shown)


pathfilters/pathfilters.go
```diff
@@ -119,11 +121,19 @@ func SDKFilter(sdk string, allowed bool) pathutil.FilterFunc {
...
 
-			for _, buildConfiguratioon := range buildConfigurations {
-				sdk, err := buildConfiguratioon.BuildSettings.String("SDKROOT")
+			for _, buildConfiguration := range buildConfigurations {
+				sdk, err := buildConfiguration.BuildSettings.String("SDKROOT")
 				if err == nil {
 					sdkMap[sdk] = true
 				}
+
+				supportedPlatformsValue, err := buildConfiguration.BuildSettings.String("SUPPORTED_PLATFORMS")
```
tothszabi, line 130: I was playing with Xcode to check if it will automatically update the `SDKROOT` value when I modify the supported platforms. And it does not do that. So I created an macOS app and then modified the supported platforms value to support every platform. The build will fail on any additional platforms and only succeed on macOS because the code does not support it officially. 

Based on this I think it would be great if we only run the extra logic when the `SDKROOT` value is `auto`:
```
if sdk != "auto" {
    continue
}
supportedPlatformsValue, err := buildConfiguration.BuildSettings.String("SUPPORTED_PLATFORMS")
...
```

godrei: Nice finding, updated the code in: https://github.com/bitrise-io/go-xcode/pull/198/commits/2d86e20eaad4c8c8dde3606ccc0a3d5192ca4694

tothszabi (approved):

PR #199, opened 2023-07-06
Add SPM support to the test command

tothszabi:
The xcode-test-mac step cannot handle the `Package.swift` file. This PR adds support for that. 

The test command relied on a supplied variable telling it if a project file is a workspace or xcproj file. This is redundant because this can be inferred from the files extension and build command builder already does it like that:
https://github.com/bitrise-io/go-xcode/blob/85e146052507bde0159551e7733c86f0067ff8b3/xcodebuild/build.go#L139-L143

This also means that I have removed the `isWorkspace` parameter from the init function which is a breaking change but only the xcode-test-mac step uses this struct from all of our steps. And I will be modifying that step after this PR and pulling in these changes. So others will not be facing this breaking change.

The other extra thing is that we need to set the working directory to where the project file is located. This is needed because xcodebuild does not have a parameter for the SPM projects. It will automatically detect it if it is run from the same folder. This is not needed for the workspaces and xcodeproj file but for consistency it will be set on those cases as well.


xcodebuild/test_test.go
```diff
@@ -151,12 +143,29 @@ func TestTestCommandModel_cmdSlice(t *testing.T) {
...
+			destination:               "",
+			generateCodeCoverage:      false,
+			disableIndexWhileBuilding: false,
+			customBuildActions:        []string{""},
+			customOptions:             []string{""},
+			want: []string{
+				"xcodebuild",
+				"-scheme",
+				"CoolLibrary",
+				"",
```
godrei, line 159: NOTE: I know that this is unrelated to the change, but don't we want to get rid of these empty command arguments?

godrei (approved):

xcodebuild/test_test.go
tothszabi: What caused is that we supplied an array of one item instead of an empty array in the test case. The logic than thought that the empty string is a valid extra parameter which needs to be included in the list of parameters. I have removed these.

godrei (approved):

PR #203, opened 2023-07-25
[BE-1124] Limit iOS simulator runtime to runtimes released for a given Xcode

lpusok:
### Context

When Xcode 14.3 and Xcode 15 (beta)  is present on the same machine, requiring the `latest` (this is the default) version of the Simulator runtime, will find iOS 17 (beta). This is undesirable as the beta Simulator can be unstable or at least not officially supported by earlier Xcodes.

In practice Xcode still works with newer runtimes, at least across the same major version (despite this table: https://developer.apple.com/support/xcode/). When there are different major versions, there is some risk of not working right.

```
Input:
- xctestrun: /var/folders/69/qgnyrbkx23dbp0p6p_b6bc9m0000gn/T/test_bundle928674358/BullsEye_FullTests_iphonesimulator16.4-arm64-x86_64.xctestrun
- destination: platform=iOS Simulator,name=iPhone 12 Pro Max
- xcodebuild_options: -parallel-testing-enabled YES
- test_repetition_mode: none
- maximum_test_repetitions: 3
- relaunch_tests_for_each_repetition: false
- BITRISE_DEPLOY_DIR: /Users/vagrant/deploy
- BITRISE_TEST_RESULT_DIR: /var/folders/69/qgnyrbkx23dbp0p6p_b6bc9m0000gn/T/test_results3063414584/test_result52921623
Creating missing device...
[11:39:06] $ xcrun "simctl" "create" "iPhone 12 Pro Max" "com.apple.CoreSimulator.SimDeviceType.iPhone-12-Pro-Max" "com.apple.CoreSimulator.SimRuntime.iOS-17-0"
Simulator device:
- name: iPhone 12 Pro Max, version: 17.0, UDID: 26EC21F3-BD7E-4734-A91C-CEBD84021316, status: Shutdown
```

Integrated in a Step:
https://github.com/bitrise-steplib/bitrise-step-xcode-test-without-building/pull/9

### DIFF
--- a/destination/xcode_runtime_support.go
+++ b/destination/xcode_runtime_support.go
@@ -1,0 +1,18 @@
+package destination
+
+func isIOSRuntimeSupportedByXcode(runtimeMajorVersion, xcodeMajorVersion int64) bool {
+	// Very simplified version of https://developer.apple.com/support/xcode/
+	// Only considering major versions for simplicity
+	var latestSupportedIOSRuntime = map[int64]int64{
+		15: 17,
+		14: 16,
+		13: 15,
+	}
+
+	latestSupportedMajorVersion, ok := latestSupportedIOSRuntime[runtimeMajorVersion]
+	if !ok {
+		return true
+	}
+
+	return latestSupportedMajorVersion >= runtimeMajorVersion
+}
--- a/destination/xcode_runtime_support.go
+++ b/destination/xcode_runtime_support.go
@@ -5,13 +5,25 @@ import (
 	"github.com/hashicorp/go-version"
 )
 
-func isIOSRuntimeSupportedByXcode(runtimeVersion *version.Version, xcodeVersion xcodeversion.Version) bool {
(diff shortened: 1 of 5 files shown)


destination/xcode_runtime_support.go
```diff
@@ -0,0 +1,45 @@
...
+
+	platformToLatestSupportedVersion, ok := xcodeVersionToSupportedRuntimes[xcodeVersion.MajorVersion]
+	if !ok {
+		return true
+	}
+
+	latestSupportedMajorVersion, ok := platformToLatestSupportedVersion[runtimePlatform]
+	if !ok {
+		return true
+	}
```
tothszabi, line 42: I like this. If something is missing (we forgetting updating the compatibility list above) then we always return with a `true`. This way we wont accidentally break something.

tothszabi (approved):

ofalvai (approved):

PR #204, opened 2023-07-27
Add support for simulator arch flag

lpusok:
### Context

Add arch flag to Simulator when booting, this allows booting Simulator in Rosetta mode.


simulator/simulator.go
```diff
@@ -22,7 +23,7 @@ const (
 type Manager interface {
 	LaunchWithGUI(simulatorID string) error
 	ResetLaunchServices() error
-	Boot(id string) error
+	Boot(device destination.Device) error
```
tothszabi, line 26: This seems like a breaking change. As I saw only the xcode-test and xcode-start-simulator steps use this so will you update both?

pgyula (approved):

PR #205, opened 2023-08-03
Added xcodecommand package

lpusok:
Moved the xcodecommand package from xcode-test (https://github.com/bitrise-steplib/steps-xcode-test/tree/ab4b14a227a9230a8164fafef22f31d8a8d7b08e/xcodecommand) to here, in preparation for adding xcbeautify support.
The xcpretty pacakge is left here, until references to it are removed.

ToDo in followup PR:
- remove v1 progress references


xcodecommand/fallbackRunner.go
```diff
@@ -0,0 +1,48 @@
...
+		fallbackRunner: NewRawCommandRunner(logger, commandFactory),
+		logger:         logger,
+	}
+}
+
+// CheckInstall checks if the wrapped Runner is available,
+// if not changes the current runner to the raw xcodebuild runner
+func (f *FallbackRunner) CheckInstall() (*version.Version, error) {
+	if f.runner == nil || f.fallbackRunner == nil {
+		panic("runner or fallback runner is nil")
```
godrei, line 29: Do we need this check? this never happens if the constructor (NewFallbackRunner) is used.
+ we don't do this check in the Run function.

If we want to not let consumers misuse FallbackRunner, we could make the struct private and the constructor could return a Runner interface.

xcodecommand/fallbackRunner.go
```diff
@@ -0,0 +1,48 @@
...
+	runner         Runner
+	fallbackRunner Runner
+	logger         log.Logger
+}
+
+// NewFallbackRunner wraps an xcbeatufy (or other) runner
+func NewFallbackRunner(runner Runner, logger log.Logger, commandFactory command.Factory) *FallbackRunner {
+	return &FallbackRunner{
+		runner:         runner,
+		fallbackRunner: NewRawCommandRunner(logger, commandFactory),
```
godrei, line 20: not sure if it make sense to create and store the fallbackrunner at this point, because if `CheckInstall` isn't failing, it never gets used.

What about creating it only when needed? (This way we would need to store the arguments needed for the NewRawCommandRunner)

xcodecommand/xcbeautify.go
```diff
@@ -0,0 +1,111 @@
...
+// CheckInstall checks if xcbeautify is on the PATH and returns its version
+func (c *XcbeautifyRunner) CheckInstall() (*version.Version, error) {
+	c.logger.Println()
+	c.logger.Infof("Checking log formatter (xcbeautify) version")
+
+	versionCmd := c.commandFactory.Create(xcbeautify, []string{"--version"}, nil)
+
+	out, err := versionCmd.RunAndReturnTrimmedOutput()
+	if err != nil {
+		if errorutil.IsExitStatusError(err) {
```
godrei, line 103: Note: we could use this approach: https://github.com/bitrise-io/go-xcode/pull/205/files#diff-fe92422579b8b9157f3742236d24f0cad571bfb1c05472ca0e7af8891b97b195R83 to get rid of errorutil.

xcodecommand/fallbackRunner.go
```diff
@@ -0,0 +1,48 @@
+package xcodecommand
+
+import (
+	"github.com/bitrise-io/go-utils/v2/command"
+	"github.com/bitrise-io/go-utils/v2/log"
+	version "github.com/hashicorp/go-version"
+)
+
+// FallbackRunner falls back to the raw xcodebuild runner in case xcbeautify (or other) is unavaialble
```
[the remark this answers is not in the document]
lpusok: Removed fallbackrunner.

xcconfig/xcconfig_test.go
```diff
@@ -33,21 +33,16 @@ func Test_WhenWritingXCConfigContent_ThenItShouldReturnFilePath(t *testing.T) {
 	if assert.NoError(t, err) {
 		assert.Equal(t, expectedPath, path)
 	}
-
-	mockPathProvider.AssertExpectations(t)
```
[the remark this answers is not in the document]
lpusok: The newly generated mocks do call AssertExpectations, as long they are crated using the New... constructor function is used.
https://github.com/bitrise-io/go-xcode/blob/f4871710a7a2bd92665c6026f416445315be54fe/xcconfig/mocks/FileManager.go#L162

xcodecommand/xcbeautify.go
lpusok: replaced.

xcodecommand/fallbackRunner.go
lpusok: This file is now removed. Raw command runner is created on demand, in this other repo: https://github.com/bitrise-steplib/steps-xcode-test/pull/234.

xcconfig/mocks/FileManager.go
```diff
@@ -1,10 +1,13 @@
-// Code generated by mockery v2.9.4. DO NOT EDIT.
```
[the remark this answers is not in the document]
lpusok: Usually I keep them in the package dir, as long as not duplicated. In this case I moved them up, as they can be useful for other packages in the future.

xcodecommand/mock_xcprettyManager.go
```diff
@@ -0,0 +1,90 @@
+// Code generated by mockery v2.13.1. DO NOT EDIT.
```
godrei, line 1: Shouldn't this be in the root mock dir too?

xcconfig/xcconfig_test.go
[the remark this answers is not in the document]
godrei: Good to know, thanks for the info!

xcodecommand/mock_xcprettyManager.go
lpusok: It is mostly used in the `xcpretty_test.go`, there is no need to mock it outside from this package, as the mock interface is used internally. There was preexisting xcpretty installation code moved and adapted in this pacakge, that is why it is exists.

godrei (approved):

PR #207, opened 2023-08-14
Xcode schemes

godrei:
### Context

Our tooling (go-xcode libs, ios steps, and project scanner) expects shared schemes to be placed in the project's shared data directory (`<project_name>.xcodeproj/xcshareddata/xcschemes`) as a `.xcscheme` file.

This expectation is not necessarily valid, by default Xcode projects are generated with a shared scheme, but this scheme isn't in the shared data dir as a scheme file. The scheme file gets generated when you modify any scheme related setting, but Xcode works well without it.

This PR updates the `xcodeproject` package's scheme management to make it more similar to how Xcode handles schemes.

Related ticket: https://bitrise.atlassian.net/browse/BIVS-2082

### Changes

**1, New scheme listing logic**

The visible schemes are the project's shared and user schemes (of the current user). Other user's schemes are not considered by Xcode.

The default schemes are not living as xcscheme files, but Xcode stores information about them under the current user's `xcschememanagement.plist`.

If the project doesn't have any visible schemes, Xcode recreates the default schemes unless the 'Autocreate schemes' option is disabled. If this option is disabled, Xcode presents a 'No schemes’ message and no actions are available.

In the case of workspaces, the logic is pretty much the same, but the 'Autocreate schemes' option is stored under the workspace and the same project level options are not considered.

**2, Missing properties added to the scheme data model**

The `skipped` and `parallelizable` fields were missing from the `TestableReference`. Because the package works with complete data models, I just added them, but they aren't used by the code.

The `shouldAutocreateTestPlan` was also added to the `TestAction`, in a follow-up PR we might update Test Plan handling, to utilize this new property.


xcodeproject/testhelper/testhelper.go
```diff
@@ -34,12 +34,17 @@ func GitCloneBranchIntoTmpDir(t *testing.T, repo string, branch string) string {
...
-	cmd := command.New("git", "clone", "-b", branch, repo, tmpDir)
-	require.NoError(t, cmd.Run())
+	GitCloneBranch(t, repo, branch, tmpDir)
 
 	return tmpDir
 }
 
+// GitCloneBranch clones a branch from a git repository into an existing directory
+func GitCloneBranch(t *testing.T, repo string, branch string, dir string) {
+	cmd := command.New("git", "clone", "-b", branch, repo, dir)
```
ofalvai, line 44: nit: it's worth adding `--depth=1` for fast tests

ofalvai (approved):

xcodeproject/xcodeproj/schemes.go
```diff
@@ -0,0 +1,253 @@
...
+	userSchemesDir, err := p.userSchemesDir()
+	if err != nil {
+		return nil, err
+	}
+	return listSchemeFilePaths(userSchemesDir)
+}
+
+func (p XcodeProj) userSchemesDir() (string, error) {
+	// <project_name>.xcodeproj/xcuserdata/<current_user>.xcuserdatad/xcschemes/
+	currentUser, err := user.Current()
```
lpusok, line 185: Does it make sense to check for the current user only, if running in a CI env?
I assume in most cases schemes would be recreated if a specific user schema would not be found, so maybe would not break for most people who relied on the previous version that did not consider the current user.

xcodeproject/testhelper/testhelper.go
godrei: added in: https://github.com/bitrise-io/go-xcode/pull/207/commits/473d0c3d31c62888580918cadf759e5428f684a4

xcodeproject/xcodeproj/schemes.go
godrei: I think we should follow how Xcode behaves, to avoid confusion.
I don't think it is too frequent, that you intentionally make a scheme not shared and then customize it.
Also, steps and tools depending on this package might be running locally too (this is probably the minority of uses, but still I think we should consider it), not just in CI.

ofalvai (approved):

xcodeproject/xcodeproj/schemes.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	log.TDebugf("%d scheme(s) found", len(schemes))
+	return schemes, nil
+}
+
+// SchemesWithAutocreateEnabled returns the schemes considered by Xcode, when opening the given project as part of a workspace.
+// SchemesWithAutocreateEnabled behaves similarly to XcodeProj.Schemes,
+// the only difference is that the 'Autocreate schemes' option is coming from the workspace settings.
+func (p XcodeProj) SchemesWithAutocreateEnabled(isAutocreateSchemesEnabled bool) ([]xcscheme.Scheme, error) {
```
lpusok, line 65: Schemes could be implemented as SchemesWithAutocreateEnabled(false).

xcodeproject/xcodeproj/schemes.go
```diff
@@ -0,0 +1,275 @@
...
+		if errors.Is(err, os.ErrNotExist) {
+			// By default 'Autocreate Schemes' is enabled
+			return true, nil
+		}
+
+		return false, err
+	}
+
+	var settings serialized.Object
+	if _, err := plist.Unmarshal(workspaceSettingsContent, &settings); err != nil {
```
lpusok, line 224: nitpick: are we ignoring directories, symlinks here?

lpusok (approved):
Great work.

PR #213, opened 2023-08-29
Improve simulator-related error messages

ofalvai:
### Context

When the destination specifier of the xcode-test step contains an invalid destination, the error messages are cryptic and not very helpful. This PR improves all the related error messages, including suggestions wherever possible.

### Changes
- Improve returned errors
- Rename a few vars and fields to make the clone easier to understand
- Tests: it's not visible in the diff, but `device_finder_test.go` tests this logic extensively

### DIFF
--- a/destination/parse.go
+++ b/destination/parse.go
@@ -164,13 +164,13 @@ func (d deviceFinder) parseDeviceList() (*deviceList, error) {
 
 		for _, deviceList := range list.Devices {
 			for _, device := range deviceList {
-				if device.IsAvailable {
-					return nil
+				if !device.IsAvailable {
+					d.logger.Warnf("device %s is unavailable: %s", device.Name, device.AvailabilityError)
 				}
 			}
 		}
 
-		return fmt.Errorf("no device is available")
+		return nil
 	}); err != nil {
 		return &deviceList{}, err
 	}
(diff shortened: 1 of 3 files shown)


destination/parse.go
```diff
@@ -164,45 +164,46 @@ func (d deviceFinder) parseDeviceList() (*deviceList, error) {
 
 		for _, deviceList := range list.Devices {
 			for _, device := range deviceList {
-				if device.IsAvailable {
```
ofalvai, line 167: Not sure what was the original idea behind this control flow, but I think it's better to loop through all devices and only print a warning for unavailable devices.

destination/parse.go
```diff
@@ -164,45 +164,46 @@ func (d deviceFinder) parseDeviceList() (*deviceList, error) {
...
-	// If the default name is required and already created will use that.
-	for _, device := range devices {
-		if device.Name == wantedDevice.Name {
+	// If the default name is required and already created, it will use that.
+	for _, device := range devicesOfRuntime {
+		if device.Name == wantedDestination.Name {
 			if !device.IsAvailable {
-				return Device{}, fmt.Errorf("device (%s) with runtime OS (%s) is unavailable: %s", wantedDevice.Name, runtime.Version, device.AvailabilityError)
+				d.logger.Warnf("device %s for %s is unavailable: %s", device.Name, runtime.Name, device.AvailabilityError)
+				continue
```
ofalvai, line 206: It's possible that another iteration of the loop will find a matching device. And even if there is no match, there is a fallback logic down below.

destination/parse.go
lpusok: The idea was to retry the simctl call if the device list is empty. This original simctl bug may be solved since. Checking the IsAvailable flag was just a case of trying to be too paranoid, I did not see any case in parctice where the device list is present but all of them are unavailable.

lpusok: I think we should remove the retry or move this check out of the retry loop, as with the changes it will not fail (trigger a retry) in the theoretical case when the device list is empty.

lpusok (approved):

PR #216, opened 2023-10-10
Add more context to errors

tothszabi:
This PR adds more context to some of the scheme related errors.

### DIFF
--- a/xcodeproject/xcworkspace/schemes.go
+++ b/xcodeproject/xcworkspace/schemes.go
@@ -26,12 +26,12 @@ func (w Workspace) Schemes() (map[string][]xcscheme.Scheme, error) {
 
 	sharedSchemes, err := w.sharedSchemes()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to read shared schemes: %w", err)
 	}
 
 	userSchemes, err := w.userSchemes()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to read user schemes: %w", err)
 	}
 
 	workspaceSchemes := append(sharedSchemes, userSchemes...)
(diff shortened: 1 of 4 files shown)


xcodeproject/xcworkspace/schemes.go
```diff
@@ -44,30 +44,30 @@ func (w Workspace) Schemes() (map[string][]xcscheme.Scheme, error) {
...
 	isAutocreateSchemesEnabled, err := w.isAutocreateSchemesEnabled()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to read the workspace autocreate scheme option: %w", err)
 	}
 
 	for _, projectLocation := range projectLocations {
 		if exist, err := pathutil.IsPathExists(projectLocation); err != nil {
-			return nil, fmt.Errorf("failed to check if project exist at: %s, error: %s", projectLocation, err)
+			return nil, fmt.Errorf("failed to check if project exist at: %s, error: %w", projectLocation, err)
```
godrei, line 57: NOTE: I suggest droping the `, error: ` suffix. Somthing like this would be more consistent with the other errors: `failed to check if project exist at %s: %w`

xcodeproject/xcworkspace/schemes.go
```diff
@@ -44,30 +44,30 @@ func (w Workspace) Schemes() (map[string][]xcscheme.Scheme, error) {
...
 		project, err := xcodeproj.Open(projectLocation)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to open project (%s): %w", projectLocation, err)
 		}
 
 		projectSchemes, err := project.SchemesWithAutocreateEnabled(isAutocreateSchemesEnabled)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to read project schemes: %w", err)
```
godrei, line 70: I suggest including the project path in the error message, so that we know which project has the issue.
I would also unify how the current project path presents in these errors, we have embedded project related errors in line 57, 65 and 70 and they all present the project path in a different way.

godrei (approved):

PR #220, opened 2023-11-23
Add visionOS Simulator as platform

ofalvai:
Add support for handling `visionOS` destinations, devices, platform lookups, etc.

Tested in https://github.com/bitrise-steplib/bitrise-step-xcode-start-simulator/pull/10

### DIFF
--- a/destination/parse.go
+++ b/destination/parse.go
@@ -306,8 +306,8 @@ func (d deviceFinder) runtimeForPlatformVersion(wantedPlatform, wantedVersion st
 			continue
 		}
 
-		// simctl reports visionOS as xrOS
-		if runtime.Platform == "xrOS" && wantedPlatform == string(VisionOS) {
+		// simctl reports visionOS as xrOS (as of Xcode 15.1 Beta 3)
+		if (runtime.Platform == "xrOS" || runtime.Platform == "visionOS") && wantedPlatform == string(VisionOS) {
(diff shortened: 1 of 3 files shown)


destination/parse.go
```diff
@@ -303,7 +303,12 @@ func (d deviceFinder) runtimeForPlatformVersion(wantedPlatform, wantedVersion st
 
 		if runtime.Platform != "" && runtime.Platform == string(wantedPlatform) {
 			runtimesOfPlatform = append(runtimesOfPlatform, runtime)
+			continue
+		}
 
+		// simctl reports visionOS as xrOS
+		if runtime.Platform == "xrOS" && wantedPlatform == string(VisionOS) {
```
tothszabi, line 310: I think this will break in one of the next Xcode update. At the moment the Vision Pro OS is called xrOS but Apple is actively renaming it to visionOS. Once that is done this if statement will not work because `runtime.Platform == "xrOS"` will be never true. 

The problem is that this will be silent breakage so I think we should already add the new expected OS name there too. Like
```
if (runtime.Platform == "xrOS" || runtime.Platform == "visionOS") && wantedPlatform == string(VisionOS) {
    ...
}
```

ofalvai: That's a cool idea! Fixed

tothszabi (approved):

PR #224, opened 2024-01-16
Spaceship retry

godrei:
This PR updates the spaceship appstoreconnectclient, to retry all commands on `503 Service Temporarily Unavailable` errors and also updates the gems used by the client to the latest versions.

The `codesign.SelectConnectionCredentials` function is also updated to print a warning if the TFA session has expired.


codesign/codesign.go
```diff
@@ -231,6 +232,12 @@ func SelectConnectionCredentials(
 			return appleauth.Credentials{}, fmt.Errorf("failed to restore Apple ID login session: %w", err)
 		}
 
+		if session != "" &&
+			bitriseConnection.AppleIDConnection.SessionExpiryDate != nil &&
+			bitriseConnection.AppleIDConnection.SessionExpiryDate.Before(time.Now()) {
+			logger.Warnf("Two-factor session has expired at: %s", bitriseConnection.AppleIDConnection.SessionExpiryDate.Format("2006-01-02 15:04"))
```
ofalvai, line 238: nit: We could also add a suggestion and a next step, like "go to your app settings (on Bitrise) and renew the connection" or a Devcenter URL.

PR #237, opened 2024-04-15
Ipa reader

godrei:
This PR fixes the issue that the Go standard library-based zip reader (`defaultRead`) can't cope with some IPA files and throws a `zip: not a valid zip file` error.

To mitigate this issue this PR adds a fallback zip reader (`dittoReader`), which can be used by the `XCArchiveReader`s and `IPAReader` if the primary `defaultRead` fails.

This is how the new zip reader and the artifacts reads are integrated: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/199


zip/ditto_reader.go
```diff
@@ -0,0 +1,93 @@
...
+	/*
+	   -x            Extract the archives given as source arguments. The format
+	                 is assumed to be CPIO, unless -k is given.  Compressed CPIO
+	                 is automatically handled.
+
+	   -k            Create or extract from a PKZip archive instead of the
+	                 default CPIO.  PKZip archives should be stored in filenames
+	                 ending in .zip.
+	*/
+	cmd := factory.Create("ditto", []string{"-x", "-k", archivePath, tmpDir}, nil)
```
tothszabi, line 46: There is a difference in the zip and ditto reader's behaviour. The ditto reader extracts the zip during initialisation while (if I am reading the zip reader's code right) it only starts extracting when you want to open a specific file from the archive. 

If we do the extraction right in the init phase then the init is not lightweight and has this hidden side effect. Which might not be a problem but it is a bit hidden. 

If you ask me right now I would do the extraction as part of an action. So in this case when the `ReadFile` method is called.

artifacts/ios_xcarchive_reader.go
```diff
@@ -1,14 +1,17 @@
...
 
 // IOSXCArchiveReader ...
 type IOSXCArchiveReader struct {
-	zipReader Reader
+	zipReader zip.ReadCloser
 }
 
 // NewIOSXCArchiveReader ...
-func NewIOSXCArchiveReader(reader Reader) IOSXCArchiveReader {
+func NewIOSXCArchiveReader(reader zip.ReadCloser) IOSXCArchiveReader {
```
tothszabi, line 14: In this comment I am just thinking out loud. I looked at the other draft PR to see how this will be used and something caught my eye. 

At the moment the caller need to supply the zip reader. First it will try the default reader, then it needs to check if the error is a zip error and also check if ditto is available and then retry it with the ditto reader. 

My question is that does the caller need to know about all of these details? Or the `IOSXCArchiveReader` and `IPAReader` struct should handle these? 

At the moment I feel like that go-xcode package should take care of it. Like if ditto is available then always use that. If not then use the default one and also maybe print a warning message that it will fail with large zip archives. 

So to sum it up to me it looks like that this init method should be like a testing focused init method and maybe we should have another one which decides what zip struct to use so the caller should not care about it.

godrei: Good suggestion, code updated in the following commits.

tothszabi (approved):

PR #241, opened 2024-08-06
[RA-3066] Add XCArchive and IPA artifact meta parsing logic

gaborpongracz:


xcodecache/swiftpm_cache.go
```diff
@@ -12,7 +12,8 @@ import (
 // SwiftPackagesStateInvalid is the partial error message printed out if swift packages cache is invalid.
 // Can be used to detect invalid state and clear the path returned by SwiftPackagesPath.
 // xcodebuild: error: Could not resolve package dependencies:
-//   The repository at [path] is invalid; try resetting package caches
+//
```
godrei, line 15: Why do we need this change? It seems the last 2 lines of the comment is an actual error message printed by xcodebuild.

metaparser/xcarchive.go
```diff
@@ -0,0 +1,76 @@
...
+	defer func() {
+		if err := reader.Close(); err != nil {
+			m.logger.Warnf("%s", err)
+		}
+	}()
+
+	xcarchiveReader := artifacts.NewXCArchiveReader(reader)
+	isMacos := xcarchiveReader.IsMacOS()
+	if isMacos {
+		m.logger.Warnf("macOS archive deployment is not supported, skipping xcarchive")
```
godrei, line 44: I think in this lib, this should be an error and the client should decide how to handle this situation.

metaparser/metaparser.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Info ...
+type Info struct {
+	AppTitle          string   `json:"app_title"`
+	BundleID          string   `json:"bundle_id"`
+	Version           string   `json:"version"`
+	BuildNumber       string   `json:"build_number"`
+	MinOSVersion      string   `json:"min_OS_version"`
+	DeviceFamilyList  []uint64 `json:"device_family_list"`
+	RawPackageContent string   `json:"-"`
```
godrei, line 27: Is this filed used?

xcodecache/swiftpm_cache.go
gaborpongracz: I've just executed a gofmt command on the whole project. So technically It wasn't me. :D

metaparser/metaparser.go
gaborpongracz: In case of Android Artifacts, this field is used during the validation process. I'm not sure, if we need this for xcode lib too.

godrei: I suggest removing this field, because it is unused and not playing a role in the JSON models we upload along the artifacts.

xcodecache/swiftpm_cache.go
godrei: Can we revert this change, to keep the original format of this example error?

metaparser/xcarchive.go
```diff
@@ -5,14 +5,21 @@ import (
 
 	"github.com/bitrise-io/go-xcode/v2/artifacts"
 	"github.com/bitrise-io/go-xcode/v2/zip"
+	"github.com/pkg/errors"
```
godrei, line 8: pls use the stdlib's error package ("errors")

metaparser/xcarchive.go
```diff
@@ -41,8 +48,7 @@ func (m *Parser) readXCArchiveDeploymentMeta(pth string) (Info, string, error) {
 	xcarchiveReader := artifacts.NewXCArchiveReader(reader)
 	isMacos := xcarchiveReader.IsMacOS()
 	if isMacos {
-		m.logger.Warnf("macOS archive deployment is not supported, skipping xcarchive")
-		return Info{}, "", nil // MacOS project is not supported, so won't be deployed.
+		return Info{}, "", MacOSProjectIsNotSupported // so won't be deployed.
```
godrei, line 51: This comment is not needed, it is up to the client what will happen with such an artifact

godrei (approved):

PR #246, opened 2024-10-02
[STEP-2061] Enterprise api

lpusok:
Adds support for using API key when using enterprise distribution method.

Changes:
- Force use of Bitrise managed signing with Enterprise (until working and tested with Xcode managed singing)
- Using the new v2 version of devportalservice
- Removed v1 appleauth.Credentials usage, moved it to devportalservice.Credentials
- Added is_enterpise_account to devportalservice so backend can send Enteprise keys
- Manual (override) connection params can also be set as Enteprise


e2e/step-manage-ios-code-signing/go.work
```diff
@@ -0,0 +1,3 @@
+go 1.22.4
+
+use ../..
```
ofalvai, line 3: 🫠

godrei (approved):

PR #247, opened 2024-10-08
Add destination to export options

godrei:
This PR adds the `destination` option to the export options.


xcodecache/swiftpm_cache_test.go
```diff
@@ -50,7 +49,7 @@ func TestCollectSwiftPackages(t *testing.T) {
 		t.Skip("Skipping integration test as -short flag is set.")
 	}
 
-	tempDir, err := ioutil.TempDir("", "")
+	tempDir, err := os.MkdirTemp("", "")
```
tothszabi, line 52: I do not know if it can be used here or not but not so long ago I learned that you can create temporary directories for the tests with `t.TempDir()` and this will be cleaned up once the test is done.

xcodeproject/xcodeproj/appiconset_test.go
```diff
@@ -128,7 +127,7 @@ func Test_appIconSetPaths(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			projectDir, err := ioutil.TempDir("", "ios-dummy-project")
+			projectDir, err := os.MkdirTemp("", "ios-dummy-project")
```
tothszabi, line 130: Same temporary folder comment as above.

xcodeproject/xcodeproj/plist.go
```diff
@@ -1,11 +1,10 @@
 package xcodeproj
 
 import (
-	"io/ioutil"
-
 	plist "github.com/bitrise-io/go-plist"
 	"github.com/bitrise-io/go-utils/fileutil"
 	"github.com/bitrise-io/go-xcode/xcodeproject/serialized"
+	"os"
```
tothszabi, line 7: Could you move this stdlib import to the top and in a separate group?

exportoptions/properties.go
```diff
@@ -165,3 +165,8 @@ const InstallerSigningCertificateKey = "installerSigningCertificate"
 
 // SigningStyleKey ...
 const SigningStyleKey = "signingStyle"
+
+// DestinationKey ...
+const DestinationKey = "destination"
+
+const DestinationExport = "export"
```
tothszabi, line 172: Will this be used as a default value starting from Xcode 15.4? Just asking because the other values in the file are named as `[key-name]Default = value`. So in this case it would be `DestinationDefault`.

certificateutil/filter_test.go
```diff
@@ -95,11 +95,10 @@ func TestFilterValidCertificateInfos(t *testing.T) {
 		},
 		{
 			name:             "2 valid, duplicated certs",
-			certificateInfos: []CertificateInfoModel{latestValidCertInfo, earlierValidCertInfo, invalidCertInfo},
+			certificateInfos: []CertificateInfoModel{latestValidCertInfo, latestValidCertInfo},
 			want: ValidCertificateInfo{
 				ValidCertificates:      []CertificateInfoModel{latestValidCertInfo},
-				InvalidCertificates:    []CertificateInfoModel{invalidCertInfo},
```
tothszabi, line 101: I do not see it from this PR but what happened to the invalid cert section here?

xcodecache/swiftpm_cache_test.go
godrei: Ohh, I didn't know about it, thanks for sharing! 
Updated in every test: https://github.com/bitrise-io/go-xcode/pull/247/commits/743c069bc48c07e36ce899d6889b4467f82bd862

xcodeproject/xcodeproj/plist.go
godrei: Fixed here: https://github.com/bitrise-io/go-xcode/pull/247/commits/2ae74667208cbbfa70df4696010105bfc3bb15ed

certificateutil/filter_test.go
godrei: This change started with the change above: https://github.com/bitrise-io/go-xcode/pull/247/files/118e1afb0d14f6d240219eeb6239184589bf76fc#diff-f623386f7f310c8d6ced0916bc0daf5dee2991b85dec4cea09d235957fd4fb12R63

After
```
earlierValidCert, privateKey, err := GenerateTestCertificate(serial, teamID, teamName, commonName, earlierValidExpiry)
```
we created a certificate info like this:
```
earlierValidCertInfo := NewCertificateInfo(*latestValidCert, earlierValidCert)
```

And some linter reported an ineffectual assignment to `privateKey`, because it wasn't used.

While in every other cert info, we used the recently created `xyzCert, privateKey`, like:
```
	invalidCert, privateKey, err := GenerateTestCertificate(serial, teamID, teamName, commonName, invalidExpiry)
	if err != nil {
		t.Errorf("init: failed to generate certificate, error: %s", err)
	}
	invalidCertInfo := NewCertificateInfo(*invalidCert, privateKey)
	t.Logf("Test certificate generated: %s", invalidCertInfo)
```

it seems it was just a typo with the `earlierValidCertInfo`.

When I fixed this lint issue, the test case started failing, so I fixed. Based on the test case title we check here if duplicates are detected.

godrei: As I expected, when the original test case was restored here: https://github.com/bitrise-io/go-xcode/pull/247/commits/bf7ca8d4bfbfbe5064c3664f5ef592ec402cbef7
The test is failing: `    --- FAIL: TestFilterValidCertificateInfos/2_valid,_duplicated_certs (0.00s)`

tothszabi (approved):

certificateutil/filter_test.go
godrei: And here the certificate filtering logic is fixed: https://github.com/bitrise-io/go-xcode/pull/247/commits/e1f107859f8c07144b77c407206d72c6e3ebf93a

PR #257, opened 2024-12-27
added testFlightInternalTestingOnly appStore option

naveen-bitrise:
Export Options now supports a `testFlightInternalTestingOnly` boolean option for app-store distribution method. 

This pull request adds that option to v1

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -1,19 +1,12 @@
-format_version: 1.0.0
+format_version: "11"
 default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
 
-app:
-  envs:
-  - ORIG_BITRISE_SOURCE_DIR: $BITRISE_SOURCE_DIR
-
 workflows:
-  ci:
+  check:
     before_run:
     - test
-    - swiftpm_cache
 
   test:
-    after_run:
-    - integration-test
     steps:
     - git::https://github.com/bitrise-steplib/steps-check.git:
         title: Lint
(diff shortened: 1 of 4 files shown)


bitrise.yml
```diff
@@ -1,19 +1,13 @@
-format_version: 1.0.0
```
godrei, line 1: The project's CI setup has been changed recently this change pulls the bitrise.yml from the master branch.

exportoptions/appstore_options.go
```diff
@@ -22,6 +22,8 @@ type AppStoreOptionsModel struct {
 	UploadSymbols bool
 	// Should Xcode manage the app's build number when uploading to App Store Connect? Defaults to YES.
 	ManageAppVersion bool
+	// TestFlightInternalTestingOnly
```
godrei, line 25: This comment is unnecessary, please remove it.

exportoptions/appstore_options.go
```diff
@@ -73,6 +75,9 @@ func (options AppStoreOptionsModel) Hash() map[string]interface{} {
 	if options.Destination != "" {
 		hash[DestinationKey] = options.Destination
 	}
+	if options.TestFlightInternalTestingOnly {
```
godrei, line 78: For the other boolean export options, we only set the given option if it has a non-default value (see `UploadBitcode` for example).
For the sake of consistency, I suggest following this pattern (constant for the default value of testFlightInternalTestingOnly [false] + only setting the option if the value is non-default).

godrei (changes requested):

godrei:
I also suggest extending the unit test to cover the new changes:
- TestNewAppStoreOptions -> should verify if the constructor creates export options with TestFlightInternalTestingOnly = false
- TestAppStoreOptionsToHash -> should verify if TestFlightInternalTestingOnly = true in the generated hash the expected key and value presents

exportoptions/appstore_options.go
naveen-bitrise: Hi @godrei,

When I used `if options.TestFlightInternalTestingOnly != TestFlightInternalTestingOnlyDefault`, I got the following linting error

```
exportoptions/appstore_options.go:79:5: S1002: should omit comparison to bool constant, can be simplified to `options.TestFlightInternalTestingOnly` (gosimple)
	if options.TestFlightInternalTestingOnly != TestFlightInternalTestingOnlyDefault {
	   ^
S1002: should omit comparison to bool constant, can be simplified to `options.TestFlightInternalTestingOnly
```

I got to a green build by commenting that off and used just `if options.TestFlightInternalTestingOnly`, let me know what I should do here.

godrei: Hi @naveen-bitrise, I updated the code to use the `if options.TestFlightInternalTestingOnly != TestFlightInternalTestingOnlyDefault` if statement. The failing linter is silenced with a `//nolint:gosimple` directive similar to the other boolean options.

I think this supports better code consistency and also expresses more that we only write non-default options to the final .plist file.

godrei (approved):

PR #260, opened 2025-03-06
Add NSunbufferedIO env to xcbeautify too

lpusok:
The xcodebuild command had this env set, but not xcbeautify; as we use go pipes not the bash command ` NSUnbufferedIO=YES xcodebuild [args] 2>&1 | xcbeautify` (reccomended by xcbeatify).

This may affect xcbeautify behaviour, and may solve some "hangs".

See: https://developer.apple.com/library/archive/technotes/tn2124/_index.html
```
NSUnbufferedIO
If set to YES, Foundation will use unbuffered I/O for stdout (stderr is unbuffered by default) 
```

### DIFF
--- a/xcodecommand/xcbeautify.go
+++ b/xcodecommand/xcbeautify.go
@@ -52,6 +52,7 @@ func (c *XcbeautifyRunner) Run(workDir string, xcodebuildArgs []string, xcbeauti
 		Stdin:  pipeReader,
 		Stdout: os.Stdout,
 		Stderr: os.Stderr,
+		Env:    xcodeCommandEnvs,
 	})
 
 	defer func() {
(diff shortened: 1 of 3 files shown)


xcodecommand/xcbeautify.go
```diff
@@ -52,6 +52,7 @@ func (c *XcbeautifyRunner) Run(workDir string, xcodebuildArgs []string, xcbeauti
 		Stdin:  pipeReader,
 		Stdout: os.Stdout,
 		Stderr: os.Stderr,
+		Env:    xcodeCommandEnvs,
```
tothszabi, line 55: Can we rename `xcodeCommandEnvs` to something like `commandEnvs` or maybe even `unbufferedIOEnv` because that is the single env it contains? Just asking because it is not only xcode now who will use the same env set.

tothszabi (approved):

xcodecommand/xcbeautify.go
lpusok: Renamed, it is more specific naming this way.

PR #263, opened 2025-03-20
[SSW-2210] Paging workarounds and improvements

lpusok:
- Added fallbacks for paging API error (invalid cursor). In case the existing paging logic fails, we use workarounds such as fetch400Profiles.
- Removed API calls for devices included in profiles, this should help with rate limits in the long term. To do this we use the UDIDs to check for profile validity, instead of the opaque ID from the API.

### DIFF
--- a/autocodesign/devportalclient/appstoreconnectclient/certificates.go
+++ b/autocodesign/devportalclient/appstoreconnectclient/certificates.go
@@ -2,9 +2,11 @@ package appstoreconnectclient
 
 import (
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"math/big"
 
+	"github.com/bitrise-io/go-utils/log"
 	"github.com/bitrise-io/go-xcode/certificateutil"
 	"github.com/bitrise-io/go-xcode/v2/autocodesign"
 	"github.com/bitrise-io/go-xcode/v2/autocodesign/devportalclient/appstoreconnect"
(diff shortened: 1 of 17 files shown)


autocodesign/devportalclient/appstoreconnectclient/certificates.go
```diff
@@ -85,13 +87,61 @@ func queryCertificatesByType(client *appstoreconnect.Client, certificateType app
...
+			certificatesByID[responseCertificate.ID] = responseCertificate
+		}
+
+		if totalCount == 0 {
+			totalCount = response.Meta.Paging.Total
+		}
+	}
+
+	if totalCount > 0 && totalCount > 400 {
+		log.Warnf("More than 400 certificates (%d) found", totalCount)
```
godrei, line 138: Maybe I would add a bit more details to this error, something like: `Unable to retrieve all certificates: more than 400 certificates available.`

godrei: This applies to all the similar warning messages.

autocodesign/localcodesignasset/profile.go
```diff
@@ -29,7 +29,7 @@ func NewProfile(info profileutil.ProvisioningProfileInfoModel, content []byte) a
 		id:             "", // only in case of Developer Portal Profiles
 		bundleID:       info.BundleID,
 		certificateIDs: nil, // only in case of Developer Portal Profiles
-		deviceIDs:      nil, // only in case of Developer Portal Profiles
+		deviceUDIDs:    nil, // (recheck this) only in case of Developer Portal Profiles
```
godrei, line 32: It might be better to add a TODO for this, which better highlights that we need to come back to this code later.

autocodesign/profiles.go
```diff
@@ -440,11 +458,25 @@ func checkProfileCertificates(profileCertificateIDs []string, certificateIDs []s
...
 
-func checkProfileDevices(profileDeviceIDs []string, deviceIDs []string) error {
-	for _, id := range deviceIDs {
-		if !sliceutil.IsStringInSlice(id, profileDeviceIDs) {
+func validDeviceUDID(udid string) string {
+	r := regexp.MustCompile("[^a-zA-Z0-9-]")
+	return r.ReplaceAllLiteralString(udid, "")
+}
+
+func normalizeDeviceUDID(udid string) string {
```
godrei, line 466: Why is this needed? Do you have an example UDID? We might add a comment with the original UDID and the transformed and why the transformation is needed.

godrei (approved):

PR #264, opened 2025-03-26
Getter for xcodebuild command arguments

lpusok:
Added CommandArgs function for the xcodebuild CommandBuilder, this removes the dependency on the v1 command package and helps with xcbeautify support integration.
The existing xcodecommand package only requires xcodebuild arguments, with this change we do no need:

```go
xcodebuildCmdWithArgs := archiveCmd.ExecCommand().Args
if len(xcodebuildCmdWithArgs) == 0 {
  panic("")
}
onlyArgs = xcodebuildCmdWithArgs[1:]
```

only:
```go
onlyArgs := archiveCmd.CommandArgs()
```

### DIFF
--- a/xcodebuild/build.go
+++ b/xcodebuild/build.go
@@ -131,9 +131,16 @@ func (c *CommandBuilder) SetTestPlan(testPlan string) *CommandBuilder {
 	return c
 }
 
-func (c *CommandBuilder) cmdSlice() []string {
+func (c CommandBuilder) cmdSlice() []string {
 	slice := []string{toolName}
-	slice = append(slice, c.actions...)
+	slice = append(slice, c.CommandArgs()...)
+
+	return slice
+}
+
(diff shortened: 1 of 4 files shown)


xcodebuild/build.go
```diff
@@ -131,9 +131,16 @@ func (c *CommandBuilder) SetTestPlan(testPlan string) *CommandBuilder {
 	return c
 }
 
-func (c *CommandBuilder) cmdSlice() []string {
+func (c CommandBuilder) cmdSlice() []string {
```
tothszabi, line 134: Goland will complain that some function defined on the regular and some on the pointer for the struct. Can you update the changed functions to also use the pointer version in the method signature here and in the `ExportCommandModel` struct?

lpusok: Updated.

tothszabi (approved):

PR #265, opened 2025-04-03
Add new export methods

lpusok:
- Added new export methods available since Xcode 15.3
- ParseMethod remains unchanged, new methods can not be used in Step inputs yet. Existing code (before export options generation) relying on exact values of Method will not break; as we will only upgrade to the new export options when generating export options.

### DIFF
--- a/exportoptions/appstore_options.go
+++ b/exportoptions/appstore_options.go
@@ -8,6 +8,7 @@ import (
 
 // AppStoreOptionsModel ...
 type AppStoreOptionsModel struct {
+	Method                             Method
 	TeamID                             string
 	BundleIDProvisioningProfileMapping map[string]string
 	SigningCertificate                 string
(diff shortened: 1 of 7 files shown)


exportoptions/appstore_options.go
```diff
@@ -26,9 +27,25 @@ type AppStoreOptionsModel struct {
 	TestFlightInternalTestingOnly bool
 }
 
-// NewAppStoreOptions ...
+// NewAppStoreOptions sets "app-store" as the export method
+// deprecated: use NewAppStoreConnectOptions instead
 func NewAppStoreOptions() AppStoreOptionsModel {
 	return AppStoreOptionsModel{
```
godrei, line 33: this could be `NewAppStoreConnectOptions(MethodAppStore)`, right?

exportoptions/properties.go
```diff
@@ -119,6 +141,20 @@ func ParseMethod(method string) (Method, error) {
 	}
 }
 
+// UpgradeExportMethod replaces the legacy export method strings with the ones available in Xcode 15.3 and later.
+func UpgradeExportMethod(method Method) Method {
```
godrei, line 145: What about adding the Xcode version to the method name? I can imagine ore changes to the method names in the future.
Something like `UpgradeToXcode15_3MethodNames`

exportoptions/properties_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestUpgradeExportMethod(t *testing.T) {
+	tests := []struct {
+		name   string
+		method Method
+		want   Method
+	}{
+		{
```
godrei, line 15: I would also test the intact method names, at least `enterprise`.

exportoptions/appstore_options.go
lpusok: Yes it is the same. It could be removed, but there is a few code using it: https://github.com/search?q=org%3Abitrise-steplib%20NewAppStoreOptions&type=code

exportoptions/properties.go
lpusok: Renamed.

exportoptions/appstore_options.go
godrei: Yeah, that make sense, I ment the function could `return NewAppStoreConnectOptions(MethodAppStore)`, instead of creating a new AppStoreOptionsModel by hand.

PR #268, opened 2025-04-15
Adding ArchiveTargetInfoProvider

lpusok:
- Added ArchveInfo. This allows to remove some legacy duplicated code in the export-xcarchive Step: https://github.com/bitrise-steplib/steps-export-xcarchive/pull/67.
- Added ReadArchiveInfoFromXcodeproject, which preserves the existing functionality when using with xcode-archive.
- Extracted some parameters of GenerateApplicationExportOptions into an Opts struct, as there were too many bool parameters that were easy to confuse. Added exportProduct and archiveInfo paramters.
- Removed some overly verbose mocking from the tests.


exportoptionsgenerator/exportoptionsgenerator.go
```diff
@@ -21,128 +21,118 @@ const (
...
-			exportOpts = addTestFlightInternalTestingOnly(exportOpts, testFlightInternalTestingOnly)
+		if opts.TestFlightInternalTestingOnly {
+			exportOpts = addTestFlightInternalTestingOnly(exportOpts, opts.TestFlightInternalTestingOnly)
 		}
 	}
 
 	return exportOpts, nil
 }
 
-func (g ExportOptionsGenerator) applicationTargetsAndEntitlements(exportMethod exportoptions.Method) (string, map[string]plistutil.PlistData, error) {
```
lpusok, line 111: Moved to targets.go, with some changes.

exportoptionsgenerator/exportoptionsgenerator.go
```diff
@@ -376,10 +366,10 @@ func addDistributionBundleIdentifierFromXcode12(exportOpts exportoptions.ExportO
 	return nil
 }
 
-func disableManagedBuildNumberFromXcode13(exportOpts exportoptions.ExportOptions) exportoptions.ExportOptions {
+func addManagedBuildNumberFromXcode13(exportOpts exportoptions.ExportOptions, isManageAppVersion bool) exportoptions.ExportOptions {
```
lpusok, line 369: Changed to support existing input (manage_version_and_build_number) in export-xcarchive.

exportoptionsgenerator/targets.go
```diff
@@ -45,32 +93,14 @@ func ArchivableApplicationTarget(xcodeProj *xcodeproj.XcodeProj, scheme *xcschem
...
-func filterApplicationBundleTargets(targets []xcodeproj.Target, exportMethod exportoptions.Method) (filteredTargets []xcodeproj.Target) {
-	fmt.Printf("Filtering %v application bundle targets", len(targets))
-
+func filterApplicationBundleTargets(targets []xcodeproj.Target) (filteredTargets []xcodeproj.Target) {
 	for _, target := range targets {
 		if !target.IsExecutableProduct() {
 			continue
 		}
 
-		// App store exports contain App Clip too. App Clip provisioning profile has to be included in export options:
```
lpusok, line 56: Moved to exportoptionsgenerator.go, so the filtering of app clips is consistent for xcode-archive and export-xcarchive.

exportoptionsgenerator/archive_info_provider.go
```diff
@@ -0,0 +1,58 @@
+package exportoptionsgenerator
```
godrei, line 1: Do we really need one more xcarchive reader? We have one reader on V1 branch, and two others on V2 already.

exportoptionsgenerator/exportoptionsgenerator.go
```diff
@@ -21,128 +21,118 @@ const (
...
 	if err != nil {
 		return nil, err
 	}
 
-	exportOpts := generateBaseExportOptions(exportMethod, xcodeVersion, uploadBitcode, compileBitcode, iCloudContainerEnvironment)
+	exportOpts := generateBaseExportOptions(exportMethod, xcodeVersion, opts.UploadBitcode, opts.CompileBitcode, iCloudContainerEnvironment)
 
 	if xcodeVersion.Major >= 12 {
-		exportOpts = addDistributionBundleIdentifierFromXcode12(exportOpts, mainTargetBundleID)
+		exportOpts = addDistributionBundleIdentifierFromXcode12(exportOpts, archiveInfo.MainBundleID)
```
godrei, line 106: I originally wanted to leave a comment, that distributionBundleID only needs to be added for non app store methods, but as I see this is already implemented inside the `addDistributionBundleIdentifierFromXcode12` function.

godrei: Also, I suggest renaming archiveInfo.MainBundleID to rchiveInfo.ProductToDistributeBundleID or something similar, which explains that you either export the main app target or the app clip target.

exportoptionsgenerator/exportoptionsgenerator.go
```diff
@@ -21,128 +21,118 @@ const (
...
 	}
 
-	mainTargetBundleID, entitlementsByBundleID, err := g.applicationTargetsAndEntitlements(exportMethod)
+	archiveInfo, err := g.archiveInfoProvider.Read()
 	if err != nil {
 		return nil, err
 	}
 
-	iCloudContainerEnvironment, err := determineIcloudContainerEnvironment(containerEnvironment, entitlementsByBundleID, exportMethod, xcodeVersion.Major)
+	// app-store-connect exports should contain App Clip too, but not other export methods:
```
godrei, line 79: I would be great to clarify this comment a bit / write a new comment to document the relationship between the distributionBundleID and the provisioningProfiles fields:
- distribution bundle ID describes which app should be exported (main app / app-clip, only valid for none app-store export)
- when distributing an app-clip, its bundle ID needs to be in the  provisioningProfiles dictionary, otherwise it needs to be removed

exportoptionsgenerator/exportoptionsgenerator.go
lpusok: Renamed MainBundleID.

exportoptionsgenerator/exportoptionsgenerator.go
lpusok: Updated comment.

exportoptionsgenerator/archive_info_provider.go
lpusok: Converted this to a simple function that uses v1 archive.
Removed one of the v2 xcarchive readers.

godrei (approved):

PR #283, opened 2025-09-02
Add a log interceptor

Badlazzor:
Singles out [Bitrise ...] logs from the rest and forwards them to stdout

### DIFF
--- a/loginterceptor/loginterceptor.go
+++ b/loginterceptor/loginterceptor.go
@@ -1,0 +1,79 @@
+package loginterceptor
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"regexp"
+	"sync"
+)
+
+// PrefixInterceptor intercept writes: if a line begins with prefix, it will be written to
+// both writers. Partial writes without newline are buffered until a newline.
+type PrefixInterceptor struct {
+	re          *regexp.Regexp
+	intercepted io.Writer
+	original    io.Writer
+
+	// internal pipe and goroutine to scan and route
+	pr *io.PipeReader
+	pw *io.PipeWriter
+
+	// close once
+	closeOnce sync.Once
+	closeErr  error
+}
+
+// NewPrefixInterceptor returns an io.WriteCloser. Writes are based on line prefix.
+func NewPrefixInterceptor(re *regexp.Regexp, intercepted, original io.Writer) *PrefixInterceptor {
+	pr, pw := io.Pipe()
+	interceptor := &PrefixInterceptor{
+		re:          re,
+		intercepted: intercepted,
+		original:    original,
+		pr:          pr,
+		pw:          pw,
+	}
+	go interceptor.run()
+	return interceptor
+}
+
+// Write implements io.Writer. It writes into an internal pipe which the interceptor goroutine consumes.
+func (i *PrefixInterceptor) Write(p []byte) (int, error) {
+	// write into pipe; pipe writer preserves order
+	return i.pw.Write(p)
+}
+
+// Close stops the interceptor and closes the pipe.
+func (i *PrefixInterceptor) Close() error {
+	i.closeOnce.Do(func() {
+		// close the writer side which causes reader side to EOF
+		i.closeErr = i.pw.Close()
+		// ensure reader is drained (run goroutine will finish)
+		_ = i.pr.Close()
+	})
+	return i.closeErr
+}
+
+// run reads lines (and partial final chunk) and writes them.
+func (i *PrefixInterceptor) run() {
+	// Use a scanner but with a large buffer to handle long lines.
+	scanner := bufio.NewScanner(i.pr)
+	const maxTokenSize = 10 * 1024 * 1024
+	buf := make([]byte, 0, 64*1024)
+	scanner.Buffer(buf, maxTokenSize)
+
+	for scanner.Scan() {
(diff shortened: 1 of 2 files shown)


loginterceptor/loginterceptor.go
```diff
@@ -0,0 +1,92 @@
...
+		i.closeErr = i.pw.Close()
+		// ensure reader is drained (run goroutine will finish)
+		_ = i.pr.Close()
+	})
+	return i.closeErr
+}
+
+// run reads lines (and partial final chunk) and writes them.
+func (i *PrefixInterceptor) run() {
+	if interceptedCloser, ok := i.intercepted.(io.Closer); ok {
```
zsolt-marta-bitrise, line 60: Can't we define them as `WriteCloser` ?

loginterceptor/loginterceptor.go
```diff
@@ -0,0 +1,92 @@
...
+
+// PrefixInterceptor intercept writes: if a line begins with prefix, it will be written to
+// both writers. Partial writes without newline are buffered until a newline.
+type PrefixInterceptor struct {
+	re          *regexp.Regexp
+	intercepted io.Writer
+	original    io.Writer
+
+	// internal pipe and goroutine to scan and route
+	pr *io.PipeReader
```
zsolt-marta-bitrise, line 20: Please don't use short and very similar (tr, pr, ti, tw, ... ) variable names in fields, that makes understanding the code really hard

loginterceptor/loginterceptor.go
Badlazzor: The MultiWriter that was already used in the xcpretty part is not a Closer unfortunately

xcpretty/xcpretty.go
```diff
@@ -53,17 +56,24 @@ func (c CommandModel) PrintableCmd() string {
...
 	pipeReader, pipeWriter := io.Pipe()
 
 	var outBuffer bytes.Buffer
 	outWriter := io.MultiWriter(&outBuffer, pipeWriter)
 
+	re := regexp.MustCompile(prefixed)
+	interceptor := loginterceptor.NewPrefixInterceptor(re, os.Stdout, outWriter)
+	defer func() {
+		if err := interceptor.Close(); err != nil {
+			loggerV1.Warnf("Failed to close log interceptor, error: %s", err)
```
zsolt-marta-bitrise, line 69: If we're using the go-utils log, can we use the same in the interceptor too?

loginterceptor/logininterceptor_test.go
```diff
@@ -0,0 +1,71 @@
...
+	sut := loginterceptor.NewPrefixInterceptor(re, iw, tw)
+
+	msg1 := "Log message without prefix\n"
+	msg2 := "[Bitrise Analytics] Log message with prefix\n"
+	msg3 := "[Bitrise Build Cache] Log message with prefix\n"
+	msg4 := "Stuff [Bitrise Build Cache] Log message without prefix\n"
+
+	go func() {
+		//nolint:errCheck
+		defer sut.Close()
```
zsolt-marta-bitrise, line 28: What happens if we just close the readers btw? Will it stop on its own?

loginterceptor/loginterceptor.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Write implements io.Writer. It writes into an internal pipe which the interceptor goroutine consumes.
+func (i *PrefixInterceptor) Write(p []byte) (int, error) {
+	return i.pw.Write(p)
+}
+
+// Close stops the interceptor and closes the pipe.
+func (i *PrefixInterceptor) Close() error {
+	i.closeOnce.Do(func() {
+		// close the writer side which causes reader side to EOF
```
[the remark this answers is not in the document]
Badlazzor: Actually we don't need the reader close after closing the writer. Closing the writer results in subsequent reads resulting in EOF. Ensuring that Close is called only when all messages are sent will be the responsibility of the calling site (just like a normal Pipe).

loginterceptor/logininterceptor_test.go
Badlazzor: The reader .Close() states: `Close closes the reader; subsequent writes to the write half of the pipe will return the error [ErrClosedPipe].` Consequently the interceptor will close with an error `writer error: io: read/write on closed pipe`. This particular test case would fail with one missing line would we replace the sut.Close() with them as the last line would not go through the scanner.
Comparing this to the behaviour of chained pipes, when a subsequent pipe closes the previous ones doesn't close themselves automatically afaik. It is true they are not stopping with an error either.

zsolt-marta-bitrise: Yeah I'm just wondering how will the infinite loop stop, so if there's nothing else to read (closed), it should close automatically, right?

Badlazzor: Yeah, I actually added a close for the `internalWriter` in the run as well. So if there is nothing to read it closes the two writers passed down to it and it closes itself. If we manually call close it closes the `internalWriter` causing the other two to close as well.

loginterceptor/loginterceptor.go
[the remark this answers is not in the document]
Badlazzor: hmm maybe I could drop the internal pipe and just write directly to the other to writers in `(i *PrefixInterceptor) Write()`

PR #286, opened 2025-09-10
[ACI-4154] Log interceptor potential deadlock fix

Badlazzor:


loginterceptor/loginterceptor.go
```diff
@@ -104,3 +99,52 @@ func (i *PrefixInterceptor) run() {
...
+		logger:  logger,
+	}
+	go writer.Run()
+	return writer
+}
+
+// Write implements io.Writer. It writes into an internal pipe which the interceptor goroutine consumes.
+func (i *NonBlockingWriter) Write(p []byte) (int, error) {
+	select {
+	case i.channel <- string(p):
```
zsolt-marta-bitrise, line 124: let's not force string conversion, we can use a []byte channel to preserve the original format

PR #291, opened 2025-09-22
Handle omitted dependencies key

realitylabs:
Newer versions of Xcode may omit the dependencies key when a target has no dependencies.


Badlazzor:
Thank you for the original fix @realitylabs — and sorry for the long wait.

Closing this in favour of #331, which:
- targets the active `v1` branch (this PR targeted `xcodeproject-v2`, where CI is failing)
- applies the same `IsKeyNotFoundError` guard but extends it to `buildPhases` as well (forward-compatibility for the same Xcode optimisation)
- has green CI

Triage tracked in STEP-2426. Credit for spotting the root cause is yours.

PR #327, opened 2026-04-24
Skip unknown entitlement keys with a warning in SyncBundleID

Badlazzor:
> **Stacked on** #326 (`audit/service-type-by-key`) → #325 (`fix/healthkit-access-entitlement`). Merge in order; this PR's diff will collapse to its own changes once the base PRs land.

## Issue

`autocodesign.Entitlement.Capability()` hard-errors with `unknown entitlement key: <key>` whenever a project's entitlements file contains a key that is not in `appstoreconnect.ServiceTypeByKey`.

`ProfileClient.SyncBundleID` iterates the entire project entitlements map and returns the first such error unwrapped. As a result, **one unrecognised key anywhere in the project kills the entire code signing phase**, even when every other entitlement is well-known and in sync with App Store Connect:

```
failed to manage code signing: failed to ensure code signing assets:
failed to ensure profiles: failed to update bundle ID capabilities:
unknown entitlement key: <key>
```

This is the failure mode behind #325 (`com.apple.developer.healthkit.access`, iOS 17.5+). Apple adds entitlement keys routinely (see #326 for a recent batch); each addition is a latent, high-blast-radius break.

## Intent

Decouple the step's stability from Apple's entitlement release cadence. An entitlement key that we have not yet classified should log a warning and be skipped, not abort the entire sync. Other entitlements on the bundle ID still get registered correctly.

## Changes

1. **Exported sentinel error** (`autocodesign.ErrUnknownEntitlementKey`) in `autocodesign/entitlement.go`. Both `Capability()` and `Equal()` now wrap it with `fmt.Errorf(\"%w: %s\", …)`. The human-readable error message is unchanged — `unknown entitlement key: <key>` — so any log-scraping stays compatible, while callers gain `errors.Is` wiring for programmatic handling.
2. **`ProfileClient.SyncBundleID`** catches the sentinel with `errors.Is`, logs a warning that identifies the offending key and points at this repo, and `continue`s the loop to process the rest of the project's entitlements instead of aborting.
3. **Regression test** (`TestCapability_UnknownKeyReturnsSentinel`) in `autocodesign/entitlement_test.go` confirms sentinel wrapping + key attribution in the error message.

`Equal()` (called from `checkBundleIDEntitlements`) is not currently reachable with an unknown key because its caller gates on `AppearsOnDeveloperPortal()`, which already returns `false` for unknown keys. The sentinel is wired there anyway for consistency — so if that gate is ever relaxed, the behavior stays defensive.

## Risk / tradeoff

**Today:** an unrecognised entitlement = every affected build fails with an opaque error.

**After this PR:** an unrecognised entitlement = a warning in the log, the bundle ID is synced with every other capability, the build proceeds. If the unknown key *did* need ASC registration, the downstream provisioning step will fail with a more targeted error (missing capability on a specific profile) that is easier to attribute than today's \"unknown entitlement key\".

In other words: we trade a guaranteed hard-break for a soft-break with better attribution. The step becomes resilient to Apple's additions until the allow-list catches up.

## Test plan

- [x] `go test ./autocodesign/...`
- [x] `go vet ./autocodesign/...`
- [x] New `TestCapability_UnknownKeyReturnsSentinel` passes
- [ ] Integration: inject a synthetic unknown key into a project's entitlements, run `xcode-archive@6.x`, confirm warning is logged and other capabilities sync correctly

### DIFF
--- a/autocodesign/devportalclient/appstoreconnectclient/profiles.go
+++ b/autocodesign/devportalclient/appstoreconnectclient/profiles.go
@@ -399,6 +399,10 @@ func (c *ProfileClient) SyncBundleID(bundleID appstoreconnect.BundleID, appEntit
 		ent := autocodesign.Entitlement{key: value}
 		cap, err := ent.Capability()
 		if err != nil {
+			if errors.Is(err, autocodesign.ErrUnknownEntitlementKey) {
+				log.Warnf("Skipping unknown entitlement key %q while syncing bundle ID capabilities. If this is a recognised Apple entitlement that should be registered on App Store Connect, please report it to bitrise-io/go-xcode so it can be added to ServiceTypeByKey.", key)
+				continue
+			}
 			return err
 		}
 		if cap == nil {
--- a/autocodesign/devportalclient/appstoreconnectclient/profiles.go
+++ b/autocodesign/devportalclient/appstoreconnectclient/profiles.go
@@ -400,6 +400,7 @@ func (c *ProfileClient) SyncBundleID(bundleID appstoreconnect.BundleID, appEntit
 		cap, err := ent.Capability()
 		if err != nil {
 			if errors.Is(err, autocodesign.ErrUnknownEntitlementKey) {
+				c.client.Tracker().TrackUnknownEntitlement(key)
 				log.Warnf("Skipping unknown entitlement key %q while syncing bundle ID capabilities. If this is a recognised Apple entitlement that should be registered on App Store Connect, please report it to bitrise-io/go-xcode so it can be added to ServiceTypeByKey.", key)
 				continue
 			}
--- a/autocodesign/entitlement.go
+++ b/autocodesign/entitlement.go
@@ -13,6 +13,14 @@ import (
 // ICloudIdentifiersEntitlementKey ...
 const ICloudIdentifiersEntitlementKey = "com.apple.developer.icloud-container-identifiers"
 
+// ErrUnknownEntitlementKey is returned by Entitlement.Capability and
+// Entitlement.Equal when an entitlement key is not present in
+// appstoreconnect.ServiceTypeByKey. Callers that iterate over a project's
+// full entitlement set (e.g. SyncBundleID) may wrap this with errors.Is to
+// skip unknown keys with a warning instead of aborting, so a single new
+// Apple-introduced key does not break the entire code signing flow.
+var ErrUnknownEntitlementKey = errors.New("unknown entitlement key")
+
 // DataProtections ...
 var DataProtections = map[string]appstoreconnect.CapabilityOptionKey{
 	"NSFileProtectionComplete":                             appstoreconnect.CompleteProtection,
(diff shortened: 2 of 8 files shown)


autocodesign/entitlement.go
```diff
@@ -13,6 +13,14 @@ import (
 // ICloudIdentifiersEntitlementKey ...
 const ICloudIdentifiersEntitlementKey = "com.apple.developer.icloud-container-identifiers"
 
+// ErrUnknownEntitlementKey is returned by Entitlement.Capability and
```
benbitrise: nit: this comment about which methods return the error is likely to become outdated. Maybe omit that part.

autocodesign/devportalclient/appstoreconnectclient/profiles.go
```diff
@@ -399,6 +399,10 @@ func (c *ProfileClient) SyncBundleID(bundleID appstoreconnect.BundleID, appEntit
 		ent := autocodesign.Entitlement{key: value}
 		cap, err := ent.Capability()
 		if err != nil {
+			if errors.Is(err, autocodesign.ErrUnknownEntitlementKey) {
+				log.Warnf("Skipping unknown entitlement key %q while syncing bundle ID capabilities. If this is a recognised Apple entitlement that should be registered on App Store Connect, please report it to bitrise-io/go-xcode so it can be added to ServiceTypeByKey.", key)
```
benbitrise, line 404: This repo tracks App Store Connect API errors (see tracker.go), and I think it would make sense to add something similar here so that we set up some alerting to get notified when the inevitable new entitlement gets added. What do you think?

autocodesign/entitlement.go
Badlazzor: Tightened the comment in fbcbb6fc520372c74514174a225e120c3689dc43 — dropped the "returned by …" phrasing, kept just what `ErrUnknownEntitlementKey` means and how callers use it.

autocodesign/devportalclient/appstoreconnectclient/profiles.go
Badlazzor: Good call — added a `TrackUnknownEntitlement(key)` method on the `Tracker` interface (with NoOp + Default impls) and wired it into `SyncBundleID` so we emit `step_appstoreconnect_unknown_entitlement` when we skip. Reachable from `ProfileClient` via a new `Client.Tracker()` accessor. Done in fbcbb6fc520372c74514174a225e120c3689dc43.

autocodesign/devportalclient/appstoreconnectclient/profiles.go
```diff
@@ -399,6 +399,11 @@ func (c *ProfileClient) SyncBundleID(bundleID appstoreconnect.BundleID, appEntit
 		ent := autocodesign.Entitlement{key: value}
 		cap, err := ent.Capability()
 		if err != nil {
+			if errors.Is(err, autocodesign.ErrUnknownEntitlementKey) {
+				c.client.Tracker().TrackUnknownEntitlement(key)
```
benbitrise, line 403: Is it feasible to add a test to confirm that the tracker is called when an unknown key is used.

autocodesign/devportalclient/appstoreconnectclient/profiles.go
Badlazzor: Also added a unit test in `21dd325` — `TestSyncBundleID_UnknownEntitlement_TracksAndSkips` asserts `TrackUnknownEntitlement` fires with the unknown key.

benbitrise (approved):

PR #331, opened 2026-05-22
fix(target): treat dependencies and buildPhases as optional keys

### LINKED ISSUE #330: parseTarget fails when dependencies key is absent (Xcode 16.2+ omits empty arrays)
## Summary

`parseTarget()` in `xcodeproject/xcodeproj/target.go` treats the `dependencies` key as mandatory. However, **Xcode 16.2+** omits the `dependencies` key entirely from `project.pbxproj` when a target has no dependencies (instead of writing `dependencies = ();`).

This causes `steps-set-xcode-build-number` (and any other consumer of `go-xcode` v1) to fail with:

```
failed to parse target with id: <ID>: key: string("dependencies") not found in: serialized.Object(...)
```

Which surfaces as the misleading error:
```
Run: failed to find build project: could not get scheme with name <Scheme> from path <Workspace>
```

## Affected targets

Any `PBXNativeTarget` or `PBXAggregateTarget` created by Xcode 16.2+ that has zero dependencies. The `dependencies` key is simply not written to the pbxproj file.

## Root cause

In [`target.go` (v1 branch)](https://github.com/bitrise-io/go-xcode/blob/v1/xcodeproject/xcodeproj/target.go):

```go
dependencyIDs, err := rawTarget.StringSlice("dependencies")
if err != nil {
    return Target{}, err  // ← hard fail on missing key
}
```

## Proposed fix

Handle `KeyNotFoundError` gracefully, the same way `productType` and `productReference` are already handled in the same function:

```go
dependencyIDs, err := rawTarget.StringSlice("dependencies")
if err != nil {
    if !serialized.IsKeyNotFoundError(err) {
        return Target{}, err
    }
}
```

The same pattern should also be applied to `buildPhases` for forward-compatibility, as Xcode may optimize that away in the future too.

## Reproduction

1. Create a new target in Xcode 16.2+ (e.g., App Extension)
2. Do NOT add any target dependencies
3. Observe `project.pbxproj` — the target entry has no `dependencies` key
4. Run any `go-xcode` v1 consumer (e.g., `steps-set-xcode-build-number`) → fails

## Workaround

Manually add `dependencies = ();` to affected targets in `project.pbxproj`. This is fragile — Xcode may strip it again on next project save.

## Environment

- Xcode 16.2, 26.2
- `go-xcode v1.0.19` (latest vendored in `steps-set-xcode-build-number v2.2.0`)
- Also affects current `v1` branch HEAD

ataetgi:
## Summary

Xcode 16.2+ omits the `dependencies` key from `project.pbxproj` when a target has no dependencies (instead of writing `dependencies = ();`). This causes `parseTarget()` to fail with a `KeyNotFoundError`.

## Changes

Apply the same `IsKeyNotFoundError` guard already used for `productType` and `productReference` to:
- `dependencies` — fixes the immediate breakage
- `buildPhases` — forward-compatibility (Xcode may optimize this away too)

## Testing

- Existing tests pass (`go test ./xcodeproject/xcodeproj/...`)
- Verified locally against a real Xcode 16.2 project with 6 affected targets

Fixes #330

### DIFF
--- a/xcodeproject/xcodeproj/target.go
+++ b/xcodeproject/xcodeproj/target.go
@@ -129,7 +129,10 @@ func parseTarget(id string, objects serialized.Object) (Target, error) {
 
 	dependencyIDs, err := rawTarget.StringSlice("dependencies")
 	if err != nil {
-		return Target{}, err
+		if !serialized.IsKeyNotFoundError(err) {
+			return Target{}, err
+		}
+		dependencyIDs = []string{}
 	}
 
 	var dependencies []TargetDependency
@@ -168,7 +171,10 @@ func parseTarget(id string, objects serialized.Object) (Target, error) {
 
 	buildPhaseIDs, err := rawTarget.StringSlice("buildPhases")
 	if err != nil {
-		return Target{}, err
+		if !serialized.IsKeyNotFoundError(err) {
+			return Target{}, err
+		}
+		buildPhaseIDs = []string{}
 	}
 
 	return Target{


Badlazzor:
**Triage notes (STEP-2426)**

- Base: `v1` (the branch consumed by step repos via `go-xcode/v1`). Confirmed: `master` no longer contains `xcodeproject/xcodeproj/target.go`, so the bug only affects the `v1` maintenance branch.
- CI on this PR is green.
- Supersedes #291 (same root cause, by @realitylabs, but targeted the dormant `xcodeproject-v2` branch and only patched `dependencies`). #291 has been closed with credit pointing here.
- Diff is strictly the `IsKeyNotFoundError` guard already used elsewhere in the file for `productType` / `productReference`. Extends the same pattern to `dependencies` and `buildPhases` — the two keys Xcode 16.2+ omits when empty.
- No new tests added; existing tests pass.

Ready for Mobile DevOps steps team review + merge.

Badlazzor (approved):
LGTM. Same `IsKeyNotFoundError` guard pattern already used for `productType` / `productReference` in this file; extending to `dependencies` and `buildPhases` is the right fix for Xcode 16.2+ omitting empty arrays.

Confirmed scope: bug only affects the active `v1` branch (master modularized `xcodeproject/` away; v2 module reuses v1 via the `go-xcode v1.3.3` go.mod dep). 34 downstream consumers identified across bitrise-io + bitrise-steplib orgs.

Will follow up with parseTarget unit tests on the xcodeproject-v2 migration work, and cut a `v1.3.4` tag once this merges.

PR #340, opened 2026-07-24
[STEP-2166] Add serialized package

tothszabi:
Migrate the v1 package to v2 with some refactoring.

In the v1 version we used named errors (`KeyNotFoundError`, `TypeCastError`) which I removed as I found better ways for them. 

The `KeyNotFoundError` was replaced by a stdlib hashmap style check:
```
name, ok := raw.String("name")
if !ok { return X{}, fmt.Errorf("...: missing 'name'") }
```
The `ok` parameters returns if the key exists or not. This is heaviest use case in the v1 package.

The `TypeCastError` is a bit more cumbersome as now:
```
v, ok := o.Bool(key)
if !ok && o.Has(key) {
    return wrong type
}
```
We need to call the `Has` function to be certain of a type mismatch. But this use case was rare and felt like we could live with it based on the previous one becoming clearer.

### DIFF
--- a/xcodeproject/serialized/serialized.go
+++ b/xcodeproject/serialized/serialized.go
@@ -13,9 +13,16 @@ package serialized
 // arbitrary values.
 type Object map[string]any
 
-// Get returns the value stored at key in o as type T. ok is false if key is
-// absent or the stored value is not a T, in which case the zero value of T is
-// returned.
+// Get returns the value stored at key in o as type T via a direct type
+// assertion. ok is false — and value is the zero value of T — if key is absent
+// or the stored value's concrete type is not exactly T.
+//
+// Because it asserts rather than converts, T must match the concrete type the
+// decoder produced. For plist/JSON input that is a scalar: string, bool, int64
+// or float64. Get does NOT handle the two composite cases that require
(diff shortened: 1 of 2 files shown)


xcodeproject/serialized/serialized.go
```diff
@@ -0,0 +1,97 @@
...
+package serialized
+
+// Object is a decoded plist/pbxproj node: a set of string keys mapped to
+// arbitrary values.
+type Object map[string]any
+
+// Get returns the value stored at key in o as type T. ok is false if key is
+// absent or the stored value is not a T, in which case the zero value of T is
+// returned.
+func Get[T any](o Object, key string) (value T, ok bool) {
```
tothszabi, line 26: Generic function seemed like a perfect fit for this one instead of duplicating the same logic for every supported type.

godrei, line 26: It is great that we have this generic Get, but I think we should highlight  that this works for primitive types only.

tothszabi: Extended the description.

godrei (approved):

PR #341, opened 2026-07-24
Migrate certificateutil to v2

benbitrise:
# ❗❗❗Notes for reviewers ❗❗❗

1. The PR description below is wordy but worthwhile for getting context before reviewing it. 
2. I don't recommend reading the diff directly. The PR is organized into several incremental commits. In order to make it easier to take in, read the PR commit by commit (Github web UI has a handy navigator for this). Each commit is small enough to easily understand the intent progressively. The first commit is a direct copy-paste of the v1 package.

## `certificateutil` v1 → v2 refactor
---

### Goal

Bring `certificateutil` from the v1 module
(`github.com/bitrise-io/go-xcode/certificateutil`) into go-xcode **v2**

While moving it, modernize rather than transcribe:

- Use modern Go (per `go.mod`, go 1.22) and current standard-library abstractions in place of roll-your-own helpers and stale external dependencies.
- Make the package **testable** — the v1 package shelled out to `security` through a package-global command constructor, which is unmockable, so almost none of it had unit tests.
- Take the migration as an opportunity to **redefine the API** around clear responsibilities, instead of copying a surface that had accumulated over years.

---
### What the v1 package did

`certificateutil` is the code-signing certificate toolkit: it parses certificates out of the various container formats Apple tooling produces (PKCS12/`.p12`, PEM, DER), models a certificate as a friendly `CertificateInfoModel`, reads the certificates installed in the macOS login Keychain (by shelling out to the`security` CLI), and offers validity/dedup filtering over the results.

It was five files with no internal go-xcode imports and a broad package-level function API.

Two structural issues stand out:

1. **Keychain access is unmockable.** Every `Installed*` function builds its command with the package-global `command.New`, so there is no seam to inject a fake — the only way to exercise them is on a real Mac with real certificates installed.
3. **Business rules are baked into the plumbing.** The "codesigning" vs "installer" distinction — a *classification* concern — is fused into distinct keychain-reading functions (`InstalledCodesigningCertificateInfos` vs `InstalledInstallerCertificateInfos`), and the same policy split is duplicated across the `*Names` / `*Certificates` / `*Infos` variants.

---

### Approach (high level)

Rather than a big-bang rewrite, the package was moved and reshaped in small, independently-reviewable commits.

1. **Land a verbatim baseline first**, so every subsequent change is a readable diff against known-good code rather than one large drop.
2. **Modernize in place** — swap stale deps for the standard library and adopt modern idioms — while the code still has its original structure, so those diffs are pure "same behavior, current Go".
4. **Separate concerns into cohesive files** *before* changing behavior: split the one grab-bag file into parsing vs keychain access, keeping moves verbatim and doing any cleanup as a distinct follow-up step. This is what made the keychain half small enough to reason about on its own.
5. **Introduce the testability seam**, injecting a `command.Factory` behind a `CertificateLister` interface — the pivotal change — and immediately add the tests it unlocks.
6. **Redefine the API last**, once the pieces were clean and covered: move classification onto the model as shared predicates, then trim the lister and delete the surface that classification made redundant.

A guiding rule throughout: **whenever code moved between files it moved verbatim, and any modification to it was a separate commit** — so a reviewer can always tell a relocation apart from a behavior change. The detailed step list is in the git history; it is intentionally not duplicated here.

**Deliberately out of scope (follow-up PR):** repointing the ~11 internal go-xcode files and the 4 direct steplib steps from v1 → v2, and deleting `autocodesign`'s now-redundant private predicates in favor of the shared `IsDevelopment` / `IsDistribution`methods.

---


certificateutil/info_model.go
```diff
@@ -0,0 +1,86 @@
...
+	SHA1Fingerprint string
+
+	Certificate x509.Certificate
+	PrivateKey  any
+}
+
+// String ...
+func (info CertificateInfoModel) String() string {
+	team := fmt.Sprintf("%s (%s)", info.TeamName, info.TeamID)
+	certInfo := fmt.Sprintf("Serial: %s, Name: %s, Team: %s, Expiry: %s", info.Serial, info.CommonName, team, info.EndDate)
```
lpusok, line 32: Could we use a pattern similar to https://github.com/bitrise-io/go-xcode/blob/master/profileutil/profile_printer.go?
We need to use logger here anyway. The printer solution removes the need to add a logger to the CertificateInfoModel struct.

As a follow-up we could check what logging do we use in steps for certs. My memory says it is a mishmash of custom Stringers and JSON loggers. Maybe there is some chance to unify it somewhat.

benbitrise: This is a very nice idea. What do you think about:

1. Merge this PR with existing CertificateModel::String()
2. Update to something like profiler_printer before migrating steps to v2
3. Only then do the downstream adoption of certicate-util v2

lpusok: Good with me

lpusok (approved):

PR #342, opened 2026-07-30
Bb/cert util printing

benbitrise:
[Previous certificate-utils v1-v2 migration PR comment](https://github.com/bitrise-io/go-xcode/pull/341#discussion_r3672626010) asked for the profile_printer.go pattern. Rendering belongs in a printer that owns the logger, not on the model.

This adds `certificate_printer` which owns every way a certificate can be printed. The consuming steps' debug blocks will print consistently and safely (not leaking private key).


profileutil/time_provider.go
```diff
@@ -1,16 +1,11 @@
...
-import "time"
+import "github.com/bitrise-io/go-xcode/v2/timeutil"
 
-// TimeProvider ...
-type TimeProvider interface {
-	Now() time.Time
-}
+// TimeProvider is an alias kept for backwards compatibility; the canonical declaration moved to
+// timeutil so certificateutil can share it.
+type TimeProvider = timeutil.TimeProvider
```
gergely-sallai: I think this is kept because some downstream tool is still using it, ones that we don't want to hunt down now, right?

If so couldn't we mark it as Deprecated, so that we at least get notified to migrate?

Or what would actually happen if we'd delete this entirely? Wouldn't we be just an import away from fixing it in the downstream usage sites? Seems like a trivial fix, but I may not see the whole picture.

certificateutil/certificate_printer.go
```diff
@@ -0,0 +1,71 @@
...
+
+// PrintableCertificate returns a verbose, indented-JSON description of the certificate, for debug
+// logging alongside profileutil's PrintableProfile.
+//
+// The certificate's raw x509 form and its private key are deliberately never included.
+func (printer *CertificatePrinter) PrintableCertificate(info CertificateInfoModel) string {
+	printable := map[string]any{}
+	printable["name"] = info.CommonName
+	printable["serial"] = info.Serial
+	printable["team"] = fmt.Sprintf("%s (%s)", info.TeamName, info.TeamID)
```
gergely-sallai, line 51: nit: double implementation on [line 33](https://github.com/bitrise-io/go-xcode/pull/342/changes#diff-e0432454e1163ea16cd7836c6467b29ca2306b8bd460aeaff55c31c3d0be958dR33), but its trivial formatting, no logic. Just caught my eyes.

Not sure if we care about this, but easy to drift overtime. But i guess the test will catch it.

certificateutil/private_key.go
```diff
@@ -0,0 +1,38 @@
...
+// String implements fmt.Stringer, masking the key.
+func (k PrivateKey) String() string {
+	if k.key == nil {
+		return ""
+	}
+	return redacted
+}
+
+// MarshalJSON implements json.Marshaler, masking the key.
+func (k PrivateKey) MarshalJSON() ([]byte, error) {
```
gergely-sallai, line 33: I know this is not easy to solve generally in Go, but its a bit fragile. If someone were to masrhal this to YAML instead of json I think it would happily write the key instead of [REDACTED].

Wild guess, but this is a common problem I think, don't we already have a solution for this somewhere in go-utils?

Maybe a wrapper type of `secret` that guards against this? We'd at least have one place to update.

gergely-sallai (dismissed):

certificateutil/private_key.go
benbitrise: Thanks. I will do a search to see how this may already be solved.

profileutil/time_provider.go
benbitrise: Yes, I should mark as deprecated if this approach toward backward compatibility is followed. 

Regarding deletion: Your intuition is correct. I kept it like this to avoid breakage where it's used in places I'm not aware of. My thought was that I could delete it, but if we're following semver strictly, then it's a major version change, which I wanted to avoid. I guess since we vendor go dependencies, we don't have to be so strict....

gergely-sallai: Yes, but we ARE doing a major version release, aren't we?
Correct me if I'm wrong, but I think this is part of v1-v2. V2 being the major version change we're doing.

benbitrise: Ah, you are right! Even though v2 has been around for a verrrry long time, it looks like it is still pre-release. 
image

certificateutil/private_key.go
benbitrise: "I think it would happily write the key" - it will not. This is because the key is unexported (lowercase). So the MarshalJSON not actually doing the important piece, rather, it is just providing the "[REDACTED]" string. Without it, we would still be safe, it just would be an empty object rather than the standard "[REDACTED]". Added additional tests to show this.

gergely-sallai: thanks for verifying!

gergely-sallai (approved):

PR #347, opened 2026-09-10
[STEP-2526] Make errorfinder streaming and tee it off logio

lpusok:
**Stacked PR 2/2** — base is `STEP-2526-logio-race` (#346, the minimal fix). This one carries the proper design on top; the diff shown here is only the streaming commit. Jira: [STEP-2526](https://bitrise.atlassian.net/browse/STEP-2526).

## What this does

#346 removed `ErrorFinder` from the runners and scans the drained raw output with `FindXcodebuildErrors`. That works, but it copies the whole log to do it, on xcbeautify/xcpretty it only sees the *filtered* log (no `[Bitrise …]` lines — review finding), and `FindXcodebuildErrors` on chunks was the design flaw in the first place: it forced error collection into the byte streams (the panic) and bisected lines at chunk boundaries. This PR moves error finding onto a streaming `Finder` that sees the complete output and assembles lines itself.

- **`errorfinder.Finder`** — an `io.Writer` form of the same line-oriented state machine (the per-line `switch` verbatim), fed incrementally. Manual `\n` splitting, no `bufio.Scanner`; lines over 1MB are *skipped*, not truncated — a truncated command line can look like an error. `FindXcodebuildErrors` is now a thin wrapper over it, and the existing suite passes unchanged.
- **`logio.SetupPipeWiring(filter, tees ...io.Writer)`** — variadic, source-compatible. Tees receive xcodebuild's **complete** output, both streams, *in front of* the filter, as one `io.MultiWriter` shared by both `exec.Cmd` fields — so `cmd.Stdout == cmd.Stderr` still holds and exec keeps one copy goroutine. That is exactly what go-utils' collector used to see. logio gains no dependency on `errorfinder`; the invariant is documented where the shared value is built.
- **Runners** pass a `Finder` as the tee and take `finder.Errors()` after `CloseFilter()`. The raw runner shares **one** `io.MultiWriter(&outBuffer, finder)` between both streams the same way. No whole-log copy, `errorCollector` off the path entirely, and error attribution is identical across the three formatters again.
- **`PrefixFilter.run()` read side** — `bufio.Reader.ReadSlice` fragment loop. The 10MB `Scanner` it replaces stopped permanently on a longer line, lost everything after it, and left exec's copy goroutine blocked on a pipe nobody read — the step hung. Fragments are routed by the anchored regexp on the first one and forwarded as they come (`bytes.Clone`d — `Sink` retains what it is given). `pipeR.Close()` in the `defer` turns any early exit into a failure rather than a hang.

The write path — `bufio.Writer`, sinks, #295's close ordering — is still untouched.

## Tests

- `Finder`: identical `Errors()` for 1/2/7/64-byte chunking vs one shot; `Errors()` repeatable; >1MB line between two errors; overlong line closes a multiline block; trailing line without newline.
- `logio`: `XcbuildStdout == XcbuildStderr` with the `==` operator (`assert.Equal`'s `DeepEqual` would accept two distinct wrappers); tee gets the complete output and both fields stay one value; `TestPipeWiring` writes to both fields; 20MB line through `PrefixFilter` (hung before); 100 lost messages don't stall the filter (hung before).
- `xcodecommand`: mock-factory runner tests (xcbeautify + raw) asserting no `ErrorFinder`, stream identity, and that `Run`'s error carries the `error:` line, a `[Bitrise Build Cache] error:` line from the other branch, and the multi-line `xcodebuild: error:` block intact — there was no runner test calling `Run()` before.
- `go test -race` green on all three packages; full `go test ./...` green.
- `BenchmarkPrefixFilterSmallWrites` (NSUnbufferedIO shape, drain inside the timed region): 2.4–2.5 µs/op, 12 allocs vs 2.5–2.6 µs/op, 16 allocs on master.

## Behaviour notes

- Multi-line `xcodebuild: error:` / `Reason:` / `Recovery suggestion:` blocks are no longer split at 32KB chunk boundaries.
- `\r\n` now passes through as-is (`ScanLines` used to drop the `\r`).
- All three runners scan the complete output, Bitrise's own lines included — same as pre-PR go-utils behaviour, and unlike #346, which on xcbeautify/xcpretty scans only the filtered raw log.

## Follow-ups (separate tickets)

`xcresult3/xcresulttool.go` shares `combinedBuffer` between two *different* MultiWriters (same class of race, not covered); go-utils `wrapOutputs` should preserve stream identity and `errorCollector` needs a mutex for every other caller; `Sink.Push` silently drops on its 1s timeout.

[STEP-2526]: https://bitrise.atlassian.net/browse/STEP-2526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/logio/prefix_filter.go
+++ b/logio/prefix_filter.go
@@ -2,22 +2,31 @@ package logio
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"regexp"
 	"sync"
 )
 
+// readBufferSize is the size of the line fragments a longer line is forwarded in.
+const readBufferSize = 64 * 1024
+
 // PrefixFilter intercept writes: when the message has a prefix that matches a
 // regexp it writes into the `Matching` sink, otherwise to the `Filtered` sink.
+//
+// PrefixFilter is not safe for concurrent use: Write buffers into a bufio.Writer, and
+// it expects a single writer. Handing the same PrefixFilter to both exec.Cmd.Stdout and
+// exec.Cmd.Stderr satisfies that, os/exec copies both streams with one goroutine then.
 //
 // Note: Callers are responsible for closing `Matching` and `Filtered` Sinks
 type PrefixFilter struct {
 	prefixRegexp *regexp.Regexp
 
 	// internal buffered middleman between xcbuild and scan
-	filterInput bufio.ReadWriter
+	filterInput *bufio.Writer
 	pipeW       *io.PipeWriter
+	pipeR       *io.PipeReader
 
 	Matching *Sink
 	Filtered io.Writer
@@ -51,7 +60,6 @@ func (p *PrefixFilter) ScannerError() <-chan error { return p.scannerError }
 //
 // Note: Callers are responsible for closing intercepted and target writers that implement io.Closer
 func NewPrefixFilter(prefixRegexp *regexp.Regexp, matching *Sink, filtered io.Writer) *PrefixFilter {
-	// This is the backing field of the bufio.ReadWriter
 	pipeR, pipeW := io.Pipe()
 	messageLost := make(chan error, 1)
 	done := make(chan struct{}, 1)
@@ -59,8 +67,9 @@ func NewPrefixFilter(prefixRegexp *regexp.Regexp, matching *Sink, filtered io.Wr
 
 	filter := &PrefixFilter{
 		prefixRegexp: prefixRegexp,
-		filterInput:  *bufio.NewReadWriter(bufio.NewReader(pipeR), bufio.NewWriter(pipeW)),
+		filterInput:  bufio.NewWriter(pipeW),
 		pipeW:        pipeW,
+		pipeR:        pipeR,
 		closeOnce:    sync.Once{},
 		messageLost:  messageLost,
 		done:         done,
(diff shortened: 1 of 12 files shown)


logio/prefix_filter.go
```diff
@@ -104,40 +114,55 @@ func (p *PrefixFilter) Close() error {
...
+				}
 			}
-		} else {
-			if _, err := p.Filtered.Write([]byte(logLine)); err != nil {
-				p.reportMessageLost(err)
+			// Copy: fragment is a view into the reader's buffer, and Sink keeps what it is given.
+			logLine := bytes.Clone(fragment)
+			if err == io.EOF {
+				// re-append newline to the partial final chunk to preserve same output format
+				logLine = append(logLine, '\n')
```
Badlazzor, line 145: This path may append stray `\n` when prior fragment was ErrBufferFull. 

Maybe `if err == io.EOF && (len(logLine) == 0 || logLine[len(logLine)-1] != '\n')`

Badlazzor (changes requested):

logio/prefix_filter.go
Badlazzor: Oh and we should add a test to cover this

PR #348, opened 2026-09-11
feat: [BIVS-3633] emit the test case identifier as a test case property

godrei:
## What

`model3.TestCase` gains an `Identifier` field, filled from the xcresult `nodeIdentifier`, and both conversion paths write it out as a test case property when the reported name does not already carry it:

```xml
<testcase name="Score is computed when the guess matches the target" classname="BullsEyeSwiftTestingTests">
  <property name="test-identifier" value="BullsEyeSwiftTestingTests/scoreIsComputedWhenGuessMatchesTarget()"></property>
</testcase>
```

A consumer that needs the identifier composes it from the class name and the test case name, so the property is only reported when the identifier differs from that composition. On an XCTest report that is never, and those reports come out byte for byte as before.

go-steputils is bumped to `v2.0.0-alpha.55`, the released version that defines the property name constant.

## Why

For a Swift Testing test written with a display name:

```swift
@Test("Score is computed when the guess matches the target")
func scoreIsComputedWhenGuessMatchesTarget()
```

xcresult reports the display string as the test case name, and the function name lives only in the node identifier. Until now the converter kept the name and dropped the identifier, so nothing downstream could name that test any more.

That breaks test quarantine: the step builds an `xcodebuild -skip-testing` filter from the reported name, the filter matches nothing, and xcodebuild ignores it without an error. The quarantined test keeps running and nobody sees why. Carrying the identifier in the report is what lets a consumer address the test exactly, and a `<testcase>` property is the encoding that survives the whole report chain with no backend change.

Part of BIVS-3633.

## Verification

- `bitrise run test` green: lint plus the full test suite.
- New model3 test case covering a display-named test case and a plain one, both carrying their node identifier.
- New `TestParseTestCase_identifierProperty`: the property is omitted when the class name and the test case name compose the identifier, emitted for a display name, omitted when there is no identifier.
- The two golden converter tests assert that the XCTest fixtures emit no identifier property at all.
- Measured on real reports. `xcresult3-flaky-with-rerun.xcresult`: 11 test cases, all of them composable, report unchanged. The fixture app's Swift Testing suite: 8 test cases, 5 carry the property, the two display-named ones and the three in a nested suite.
- End to end against the fixture app: ran the `UnitTests` plan on Xcode 26.6 and converted the xcresult with this branch. The display-named test case carries `BullsEyeSwiftTestingTests/scoreIsComputedWhenGuessMatchesTarget()`.

## Notes for the reviewer

- The value is the node identifier as xcresult reports it, with no test target prefix. The consumer adds the target, which it already gets from the suite name.
- The legacy `genTestCase` path has no test coverage in this repo, so its change is verified by reading only. It reuses `test.Identifier.Value`, the same value the function already splits for the class name.


testresult/xcresult3/model3/conversion.go
```diff
@@ -209,10 +209,11 @@ func extractTestCase(testNode TestNode, customNodeIdentifier, customName, fallba
...
 
 	return TestCase{
-		Name:      name,
-		ClassName: className,
-		Time:      extractDuration(testNode.Duration),
-		Result:    testNode.Result,
-		Message:   message,
+		Name:       name,
+		ClassName:  className,
+		Identifier: nodeIdentifier,
```
godrei, line 214: Do we only set `Identifier` if needed (e.g.: TestCase.name != nodeIdentifier, or with other words: if display name is set for the test case) ?

godrei: Good catch. On the XCTest fixture all 11 identifiers were exactly `classname/name` and the property grew the report by 57%, so it is now emitted only when `identifier != className + "/" + name`.

That condition is the inverse of the consumer fallback, which composes the same string from the class name and the test case name, so the case we omit is always the one the consumer can rebuild.

In the sample app 5 of 8 test cases keep the property: the two display-named ones and the three in a nested suite. Plain Swift Testing cases carry nothing now, same as XCTest.

tothszabi (dismissed):

tothszabi (approved):

PR #349, opened 2026-09-18
[SSW-3174] locate the Simulator GUI on Xcode 27 (DeviceHub.app)

pgyula:
Xcode 27 removed Simulator.app from the Developer Directory. 
The Simulator GUI still exists, as Xcode.app/Contents/Applications/DeviceHub.app, but LaunchWithGUI() used the old hard-coded path unconditionally, so `open` failed and the call always errored out. That is what makes `headless_mode: no` unusable on Xcode 27 stacks.

findSimulatorGUIApp() now looks for Simulator.app first and falls back to DeviceHub.app, so behaviour on Xcode 26 and earlier is unchanged. 
DeviceHub ignores Simulator.app's -CurrentDeviceUDID argument - it accepts it and exits 0 without booting anything - so boot the device explicitly on that path. DeviceHub tracks CoreSimulator state live, so boot and open can happen in either order, and Boot() is a no-op when the device is already booted.


simulator/simulator.go
```diff
@@ -54,14 +62,48 @@ func (m manager) getSimulatorAppAbsolutePath() (string, error) {
...
+		return guiApp{}, err
+	}
+
+	if _, err := os.Stat(simulatorAppPath); err == nil {
+		return guiApp{path: simulatorAppPath}, nil
+	}
+
+	// simulatorAppPath is <developer dir>/Applications/Simulator.app, so the Xcode.app/Contents
+	// directory that holds DeviceHub.app is two levels up from the developer directory.
+	deviceHubAppPath := filepath.Clean(filepath.Join(simulatorAppPath, "..", "..", "..", "Applications", "DeviceHub.app"))
```
gergely-sallai, line 82: I'd be more comfortable with relying on an explicit new `deviceHubAppPath` state rather than doing the `"..", "..", ".."` gymnastics and thus having this tied to `simulatorAppPath`'s value.

Reasoning:
This will stay with us, this is the new app, we should be able to handle it "natively" and separately from the Simulator.app's location. This is too fragile for me.

gergely-sallai (changes requested):

simulator/simulator.go
```diff
@@ -54,14 +62,48 @@ func (m manager) getSimulatorAppAbsolutePath() (string, error) {
 	return filepath.Join(xcodeDevDirPath, "Applications", "Simulator.app"), nil
 }
 
+// findSimulatorGUIApp locates the GUI application that displays Simulators.
+//
+// Up to and including Xcode 26 this is Simulator.app inside the Xcode Developer Directory.
+// Xcode 27 removed Simulator.app entirely and replaced it with DeviceHub.app, which lives next
+// to the Developer Directory in Xcode.app/Contents/Applications.
```
ofalvai: There is no need to document implementation details in such a verbose way, please clean up these Claude-isms.

simulator/simulator.go
ofalvai: Agree, we are hardcoding implementation details anyway, let's do it the simplest way possible.

ofalvai:
FYI @pgyula @gergely-sallai I initially wanted to suggest not hardcoding any of these 2 `.app` paths and just rely on macOS and its `open` utility (these apps are indexed by Spotlight). This was a dead-end as Spotlight doesn't understand what it means to have multiple Xcodes installed side-by-side and have one actively selected Xcode (via `xcode-select`). Spotlight lists all available `Simulator.app`s and `Device Hub.app`s from all Xcodes, so `open -a "Device Hub"` would open a random one.

gergely-sallai (approved):

ofalvai (approved):

